PR #4, opened 2023-08-01
 Port test_schedule_compute.py

En-tropy:
Port test_schedule_compute.py except tests with index computation

### DIFF
--- a/tests/test_schedule_compute.py
+++ b/tests/test_schedule_compute.py
@@ -1,7 +1,6 @@
 # Copyright Allo authors. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-import sys
 import allo
 import numpy as np
 import pytest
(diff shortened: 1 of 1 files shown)


tests/test_schedule_compute.py
```diff
@@ -0,0 +1,229 @@
...
+def test_pipeline():
+    def pipeline(A: int32[10, 20], B: int32[10, 20]) -> int32[10, 20]:
+        C: int32[10, 20] = 0
+        for i, j in allo.grid(10, 20):
+            C[i, j] = A[i, j] + B[i, j]
+        return C
+
+    s = allo.customize(pipeline)
+    initiation_interval = 4
+    s.pipeline("i", initiation_interval)
```
chhzh123: No need to create an intermediate variable here, could you change it to `s.pipeline("i", initiation_interval=4)`? Similar for other test cases.

PR #11, opened 2023-08-06
[Builder] Add allo.matmul op supported by linalg

Zhichenzzz:
support tensor matmul and add test cases.

### DIFF
--- a/allo/ir/builder.py
+++ b/allo/ir/builder.py
@@ -409,10 +409,28 @@ class ASTTransformer(Builder):
         raise RuntimeError("Unsupported store")
 
     @staticmethod
+    def build_Matmul(ctx, node):
+        ip = ctx.get_ip()
+        # pylint: disable=unexpected-keyword-arg
+        with ip:
+            outputs = linalg_d.matmul(
+                ctx.buffers[node.value.args[0].id].result,
+                ctx.buffers[node.value.args[1].id].result,
+                outs=[ctx.buffers[node.targets[0].id].result],
+            )
+        return outputs
+
+    @staticmethod
     def build_Assign(ctx, node):
         # Compute RHS
         if isinstance(node.value, ast.Name):  # scalar
             rhs = ctx.buffers[node.value.id]
+        elif (
+            isinstance(node.value, ast.Call)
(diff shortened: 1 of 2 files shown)


allo/ir/builder.py
```diff
@@ -408,11 +408,29 @@ def build_store(ctx, node, val):
...
     def build_Assign(ctx, node):
         # Compute RHS
         if isinstance(node.value, ast.Name):  # scalar
             rhs = ctx.buffers[node.value.id]
+        elif (
+            isinstance(node.value, ast.Call)
+            and isinstance(node.value.func, ast.Attribute)
+            and (node.value.func.attr == "matmul")
+        ):
+            return ASTTransformer.build_Matmul(ctx, node)
```
chhzh123, line 433: You need to dispatch the function call in `build_Call` instead of building it here. It will cause problems for cases like this:

```python
D = allo.matmul(allo.matmul(A, B), C)
# or later we support broadcasting
D = 1 + allo.matmul(A, B)
```

If you build it in `Assign`, you lose the functionality of recursion.

tests/test_parser.py
```diff
@@ -824,6 +824,46 @@ def kernel() -> float32[M, N]:
...
+        C: float32[M, N] = 0.0
+        C = allo.matmul(A, B)
+        return C
+
+    s = allo.customize(kernel)
+    f = s.build()
+    np_2 = np.zeros((M, N), dtype="float32")
+    np_2 = f(np_0, np_1)
+    np.testing.assert_allclose(np_2, np_0 @ np_1, atol=1e-4)
+    print(s.module)
```
chhzh123, line 864: Can you try other cases instead of changing the data types? For example, you can test cascading two `matmul`s together just like what I mentioned above.

Zhichenzzz: Okay, I will improve it.

allo/ir/builder.py
Zhichenzzz: Okay, I've finished this func in #12 
I'm sorry that I am unfamiliar with using PR or branches, which may bring some trouble to you.

PR #14, opened 2023-08-08
tensortype and tensorgenerate

mmengjiadai:
able to input data as tensors and define new arguments as tensors

### DIFF
--- a/tests/test_tensor.py
+++ b/tests/test_tensor.py
@@ -6,49 +6,42 @@ from allo.ir.types import int32
 import numpy as np
 
 
-def same(A: int32[32, 1]) -> int32[32, 1]:
-    return A
+def test_same():
+    def same(A: int32[32, 32]) -> int32[32, 32]:
+        return A
+
+    s = allo.customize(same, enable_tensor=True)
+    print(s.module)
 
 
-def outzero() -> int32:
-    C: int32 = 0
-    return C
+def test_same_scalar():
+    def same_scalar(A: int32) -> int32:
+        return A
+
+    s = allo.customize(same_scalar, enable_tensor=True)
+    print(s.module)
 
 
-# test_td_outzero = """
-# func.func @outzero() -> tensor<i32> {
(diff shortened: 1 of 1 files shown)


allo/ir/builder.py
```diff
@@ -601,30 +604,43 @@ def build_AnnAssign(ctx, node):
...
+                ip = ctx.pop_ip()
+                ctx.buffers[node.target.id] = tensorgen_op
         elif isinstance(type_hint, ast.Name):
             type_str = type_hint.id
             if type_str in ctx.global_vars:
                 type_str = str(ctx.global_vars[type_str])
+            if ctx.enable_tensor is False:
+                # TODO: figure out why zero-shape cannot work
+                ctx.buffers[node.target.id] = MockScalar(node.target.id, type_str, ctx)
+                ASTTransformer.build_store(ctx, node.target, rhs)
```
chhzh123: Did you mistakenly copy the code? It is the same as L644-L647:
https://github.com/cornell-zhang/allo/blob/bbbb1c612dd6ed152cc561abcecf1f6958a8a9d7/allo/ir/builder.py#L644-L647

allo/ir/builder.py
```diff
@@ -665,20 +681,26 @@ def build_type(type_hint):
                     for x in elts
                 ]
                 ele_type = get_mlir_type(type_str)
-                memref_type = MemRefType.get(shape, ele_type)
+                if enable_tensor is False:
```
chhzh123: Nit: `if not enable_tensor`

allo/ir/builder.py
```diff
@@ -648,7 +664,7 @@ def build_FunctionDef(ctx, node):
         input_typehints = []
         arg_names = []
 
-        def build_type(type_hint):
+        def build_type(type_hint, enable_tensor=False):
```
chhzh123: I think you don't need to pass in `enable_tensor` here as an argument? You can just call `ctx.enable_tensor` inside the subfunction.

PR #18, opened 2023-08-12
[ir] Add matrix math Ops and softmax(can not lower) Op

Zhichenzzz:
## Description ##
This PR adds the matrix exponent and softmax frontend feature to support the backend function allo.exp(), however, there are some problems when lowering softmax func to llvm , and provides more tests.

```python
def build_linalgOp(ctx, attr, new_args):
```

### Changes and examples ###

- [Frontend] Support using the function of `allo.exp()` like this:

``` python
def kernel(A: float32[M, K], B: float32[K, M]) -> float32[M, M]:
    D = allo.matmul(A, B)
    C = allo.exp(D)
    return C
```
The generated IR:

``` mlir
module {
  func.func @kernel(%arg0: memref<10x15xf32>, %arg1: memref<15x10xf32>) -> memref<10x10xf32> {
    %cst = arith.constant 0.000000e+00 : f32
    %alloc = memref.alloc() : memref<10x10xf32>
    affine.for %arg2 = 0 to 10 {
      affine.for %arg3 = 0 to 10 {
        affine.store %cst, %alloc[%arg2, %arg3] : memref<10x10xf32>
      }
    }
    affine.for %arg2 = 0 to 10 {
      affine.for %arg3 = 0 to 10 {
        affine.for %arg4 = 0 to 15 {
          %0 = affine.load %arg0[%arg2, %arg4] : memref<10x15xf32>
          %1 = affine.load %arg1[%arg4, %arg3] : memref<15x10xf32>
          %2 = affine.load %alloc[%arg2, %arg3] : memref<10x10xf32>
          %3 = arith.mulf %0, %1 : f32
          %4 = arith.addf %2, %3 : f32
          affine.store %4, %alloc[%arg2, %arg3] : memref<10x10xf32>
        }
      }
    }
    %alloc_0 = memref.alloc() : memref<10x10xf32>
    affine.for %arg2 = 0 to 10 {
      affine.for %arg3 = 0 to 10 {
        affine.store %cst, %alloc_0[%arg2, %arg3] : memref<10x10xf32>
      }
    }
    affine.for %arg2 = 0 to 10 {
      affine.for %arg3 = 0 to 10 {
        %0 = affine.load %alloc[%arg2, %arg3] : memref<10x10xf32>
        %1 = math.exp %0 : f32
        affine.store %1, %alloc_0[%arg2, %arg3] : memref<10x10xf32>
      }
    }
    return %alloc_0 : memref<10x10xf32>
  }
}
```
- [Test] Add some corresponding test cases.
### Problems ###

- [Frontend] The function of allo.softmax() can customize successfully but it can not be lowered to llvm:
```python
def kernel(A: float32[M, K]) -> float32[M, K]:
    outs = allo.softmax(A)
    return outs
```
``` mlir
module {
  func.func @kernel(%arg0: memref<10x15xf32>) -> memref<10x15xf32> {
    %alloc = memref.alloc() : memref<10x15xf32>
    %cst = arith.constant 0.000000e+00 : f32
    linalg.fill ins(%cst : f32) outs(%alloc : memref<10x15xf32>)
    linalg.softmax dimension(1) ins(%arg0 : memref<10x15xf32>) outs(%alloc : memref<10x15xf32>)
    return %alloc : memref<10x15xf32>
  }
}
```

After discussing with Hongzheng, we guessed that developers haven't written the lowering pass for linalg yet. So this function can not build now. Please see [here](https://reviews.llvm.org/D153422).

## Checklist ##

- [x] Code is well-documented

### DIFF
--- a/allo/ir/builder.py
+++ b/allo/ir/builder.py
@@ -953,11 +953,20 @@ class ASTTransformer(Builder):
     @staticmethod
     def build_linalgOp(ctx, attr, new_args):
         ip = ctx.get_ip()
-        if attr in {"exp", "softmax", "abs", "log"}:
+        if attr in {"exp", "softmax", "abs", "log", "add", "sub", "div"}:
             dtype = ShapedType(new_args[0].type).element_type
             argshape = ShapedType(new_args[0].type).shape
             shape = argshape
-        elif attr in {"matmul", "bmm", "add", "sub", "div"}:
+            if attr in {"add", "sub", "div"}:
+                if (
+                    ShapedType(new_args[0].type).shape
+                    != ShapedType(new_args[1].type).shape
+                ):
+                    raise RuntimeError(
+                        "Only support element-wise operation of two inputs with the same shape"
+                    )
+                shape = argshape
+        elif attr in {"matmul", "bmm"}:
             # matrix shape
             dtype = ShapedType(new_args[0].type).element_type
             argAshape = ShapedType(new_args[0].type).shape
@@ -968,22 +977,24 @@ class ASTTransformer(Builder):
                         "Only support matrix multiplication of two 2D inputs"
                     )
                 if argAshape[1] != argBshape[0]:
-                    raise RuntimeError("Inputs must follow the matmul rule")
+                    raise RuntimeError(
+                        "The second dimension of the first input and the first dimension of the second input must be the same"
+                    )
                 shape = (argAshape[0], argBshape[1])
             if attr == "bmm":
                 if len(argAshape) != 3 or len(argBshape) != 3:
                     raise RuntimeError(
                         "Only support batched matrix multiplication of two 3D inputs"
                     )
-                if argAshape[2] != argBshape[1] or argAshape[0] != argBshape[0]:
-                    raise RuntimeError("Inputs must follow the bmm rule")
-                shape = (argAshape[0], argAshape[1], argBshape[2])
-            if attr in {"add", "sub", "div"}:
-                if argAshape != argBshape:
+                if argAshape[2] != argBshape[1]:
                     raise RuntimeError(
-                        "Only support element-wise operation of two inputs with the same shape"
+                        "The third dimension of the first input and the second dimension of the second input must be the same"
                     )
-                shape = argAshape
+                if argAshape[0] != argBshape[0]:
+                    raise RuntimeError(
+                        "The first dimension of the first input and the first dimension of the second input must be the same"
+                    )
+                shape = (argAshape[0], argAshape[1], argBshape[2])
         else:
             raise RuntimeError("Unsupported operation")
         with ip:


allo/ir/builder.py
```diff
@@ -935,26 +935,30 @@ def build_Call(ctx, node):
...
-                    "Only support batched matrix multiplication of two 3D inputs"
-                )
-            shape = (argAshape[0], argAshape[1], argBshape[2])
-
-        # pylint: disable=unexpected-keyword-arg
+        if attr in {"exp_m", "softmax"}:
+            dtype = ShapedType(new_args[0].type).element_type
+            argshape = ShapedType(new_args[0].type).shape
+            shape = argshape
+        if attr in {"matmul", "bmm"}:
```
chhzh123: What if the `attr` is not included in those sets? Better to raise a RuntimeError in that case

allo/ir/builder.py
```diff
@@ -974,6 +978,19 @@ def build_Matmul(ctx, attr, new_args):
                     new_args[1],
                     outs=[alloc_op],
                 )
+            if attr == "exp_m":
+                linalg_d.exp(
+                    new_args[0],
+                    outs=[alloc_op],
+                )
+            # TODO: softmax failed to lower
```
chhzh123:
```
# TODO: failed to lower to LLVM, see https://reviews.llvm.org/D153422
```

Zhichenzzz: When I added the [convert-linalg-to-affine-loops](https://mlir.llvm.org/docs/Passes/#-convert-linalg-to-affine-loops) pass or enabled lower_linalg, it still generated these errors:
```html
loc("-":10:5): error: failed to legalize operation 'linalg.softmax'
loc("-":2:3): error: cannot be converted to LLVM IR: missing `LLVMTranslationDialectInterface` registration for dialect for op: func.func
```

tests/test_linalg.py
```diff
@@ -140,10 +140,47 @@ def kernel(A: int32[M, N, K], B: int32[M, K, N]) -> int32[M, N, K]:
...
+    K = 15
+    A = np.float32(np.random.uniform(size=(M, K)))
+
+    def kernel(A: float32[M, K]) -> float32[M, K]:
+        outs = allo.softmax(A)
+        return outs
+
+    s = allo.customize(kernel)
+    print(s.module)
+    f = s.build()
```
chhzh123: I think instead of skipping this test case. It would be better to use pytest to capture the error here.

allo/ir/builder.py
```diff
@@ -915,65 +916,101 @@ def build_Call(ctx, node):
...
-                    new_args[0],
-                    new_args[1],
-                    outs=[alloc_op],
+            if attr in {"matmul", "bmm", "add", "sub", "div"}:
+                {
+                    "matmul": linalg_d.matmul,
+                    "bmm": linalg_d.batch_matmul,
+                    "add": linalg_d.add,
+                    "sub": linalg_d.sub,
+                    "div": linalg_d.div,
```
chhzh123, line 1013: Just keep in mind that we want to overload these operations instead of letting users explicitly call them. For example, prefer `A + B` rather than `allo.add(A, B)` if `A` and `B` are both tensors. You can add a TODO here and fix it in the next PR.

tests/test_linalg.py
```diff
@@ -140,10 +140,50 @@ def kernel(A: int32[M, N, K], B: int32[M, K, N]) -> int32[M, N, K]:
...
+
+    def kernel(A: float32[M, K]) -> float32[M, K]:
+        outs = allo.softmax(A)
+        return outs
+
+    s = allo.customize(kernel)
+    print(s.module)
+    with pytest.raises(RuntimeError) as excinfo:
+        f = s.build()
+    assert "Failure while creating the ExecutionEngine." in str(excinfo.value)
```
chhzh123: I'm not sure how this line works. It seems the indention is not correct. I would suggest just removing it.

allo/ir/builder.py
```diff
@@ -915,65 +916,101 @@ def build_Call(ctx, node):
...
-            "sqrt": math_d.SqrtOp,
-            "sin": math_d.SinOp,
-            "cos": math_d.CosOp,
-            "tan": math_d.TanOp,
-            "tanh": math_d.TanhOp,
-            "power": math_d.PowFOp,
-        }.get(node.func.attr)
         new_args = [stmt.result for stmt in build_stmts(ctx, node.args)]
-        return opcls(*new_args, ip=ctx.get_ip())
+        # pylint: disable=no-else-return
```
chhzh123: Should fix this one.

allo/ir/builder.py
```diff
@@ -915,65 +916,101 @@ def build_Call(ctx, node):
...
+        if attr in {"exp", "softmax", "abs", "log"}:
+            dtype = ShapedType(new_args[0].type).element_type
+            argshape = ShapedType(new_args[0].type).shape
+            shape = argshape
+        elif attr in {"matmul", "bmm", "add", "sub", "div"}:
+            # matrix shape
+            dtype = ShapedType(new_args[0].type).element_type
+            argAshape = ShapedType(new_args[0].type).shape
+            argBshape = ShapedType(new_args[1].type).shape
+            shape = (argAshape[0], argAshape[1])
```
chhzh123: A redundant line?

Zhichenzzz: The one above is a unary operation, and the one below is a binary operation. So, the shape of result depends on different ways

Zhichenzzz: I will delete the redundant line.

chhzh123: Okay, I kind of understand what you are doing here, but the coding style is not clear, which makes me think of the shape is overwritten in L972 and L978. It would be better to explicitly create an `if` branch for `add`, `sub`, and `div` since they are elementwise-binary op.

Zhichenzzz:
``` python
        if attr in {"exp", "softmax", "abs", "log", "add", "sub", "div"}:
            dtype = ShapedType(new_args[0].type).element_type
            argshape = ShapedType(new_args[0].type).shape
            shape = argshape
        elif attr in {"matmul", "bmm"}:
            # matrix shape
            dtype = ShapedType(new_args[0].type).element_type
            argAshape = ShapedType(new_args[0].type).shape
            argBshape = ShapedType(new_args[1].type).shape
            if attr == "matmul":
                if len(argAshape) != 2 or len(argBshape) != 2:
                    raise RuntimeError(
                        "Only support matrix multiplication of two 2D inputs"
                    )
                shape = (argAshape[0], argBshape[1])
            if attr == "bmm":
                if len(argAshape) != 3 or len(argBshape) != 3:
                    raise RuntimeError(
                        "Only support batched matrix multiplication of two 3D inputs"
                    )
                shape = (argAshape[0], argAshape[1], argBshape[2])

```

I think we can merge arithmetic into `exp` `softmax` since their two parameters have the same shape. Btw, do I need to check if the shapes of two parameters are the same as well as in `matmal/bmm` (fulfill the matrix multiplication rules)?

chhzh123: This also looks good. In general, we need to do the checking for the shape. I think you can add assertions for both cases.

allo/ir/builder.py
```diff
@@ -915,65 +916,109 @@ def build_Call(ctx, node):
...
+            dtype = ShapedType(new_args[0].type).element_type
+            argAshape = ShapedType(new_args[0].type).shape
+            argBshape = ShapedType(new_args[1].type).shape
+            if attr == "matmul":
+                if len(argAshape) != 2 or len(argBshape) != 2:
+                    raise RuntimeError(
+                        "Only support matrix multiplication of two 2D inputs"
+                    )
+                if argAshape[1] != argBshape[0]:
+                    raise RuntimeError("Inputs must follow the matmul rule")
```
chhzh123: It would be better to give more information about the shape. Suppose you are the user, you may get confused about "the matmul rule". We should explicitly inform them of what causes the error and how they can fix it. A good error message from PyTorch is like:
https://github.com/pytorch/pytorch/blob/6af6b8f728426fb7551630e28148c0017fa501bc/aten/src/ATen/native/layer_norm.h#L24-L29

PR #42, opened 2023-08-22
[IR]extract and insert tensor element and slice

mmengjiadai:
## Description ##
This PR enables extracting and inserting slices and elements of tensors with Python syntax when using Tensor Dialect.

### Problems ###
Aiming to implement python syntax for slicing, this PR supports the following features ():
- Extract slices from tensors with arbitrary dimensions using syntax `Array[lower: upper: step]`. For example, extract A[1:2:1, 2:4:2] from array A[6, 6] 
- Extract element from tensors with arbitrary dimensions using syntax `Array[index]`. For example, `A[1, 1]`
- Similar to extract, insert value to slice or element by directly assigning new value. For example, insert tensor B[2] to slice of tensor A[6] by `A[1:3] = B`
- Just like python, lower, upper, or step may be omitted. For example, when extracting from 1D tensor A[3], slice A[:] is the same as A[0:3:1]

### Proposed Solutions ###
- Added build_ExtSlice() and build_Index() to get arguments for extract and insert ops.
- Added `tensor_d.ExtractOp` and `tensor_d.ExtractSliceOp` to `build_subscript()` in builder.py.
- Added `tensor_d.InsertOp` and `tensor_d. InsertSliceOp` to `build_store()` in builder.py.
- Store new value of tensor in `ctx.buffers` in `build_Assign()`.
- Add pass `expand-strided-metadata` in LLVMModule() in llvm.py

### Examples ###
```
def slice(A: int32[6, 6]) -> int32[6, 6]:
        B: int32[2, 3] = 0
        B[0, 0] = 1
        A[0:2, 0:3] = B
        return A
```
We can define a tensor in the function, insert values to an extracted element, and insert it to another tensor.
## Checklist ##

- [x] Code is well-documented

### DIFF
--- a/allo/backend/llvm.py
+++ b/allo/backend/llvm.py
@@ -366,7 +366,6 @@ class LLVMModule:
                 "func.func(convert-linalg-to-affine-loops),lower-affine)"
             )
             pm.run(self.module.operation)
-            print(self.module)
             # Attach necessary attributes
             func = find_func_in_module(self.module, top_func_name)
             if func is None:
--- a/allo/ir/builder.py
+++ b/allo/ir/builder.py
@@ -756,10 +756,10 @@ class ASTTransformer(ASTBuilder):
         static_offsets = []
         static_sizes = []
         static_strides = []
-        for index, shape in zip(slices, in_shape):
+        for index, size in zip(slices, in_shape):
             if isinstance(index, ast.Slice):
                 lower = 0 if index.lower is None else index.lower.value
-                upper = shape if index.upper is None else index.upper.value
+                upper = size if index.upper is None else index.upper.value
                 if index.step is None:
                     step = 1
                 elif isinstance(index.step, ast.Constant):
@@ -772,7 +772,7 @@ class ASTTransformer(ASTBuilder):
                 step = 1
             if lower < 0 or upper < 0:
                 raise RuntimeError("Unsupported negative index")
-            if lower > shape or upper > shape:
+            if lower > size or upper > size:
                 raise RuntimeError("Index out of range")
             if step <= 0:
                 raise RuntimeError("Unsupported negative step")
@@ -836,6 +836,7 @@ class ASTTransformer(ASTBuilder):
                         ip=ctx.get_ip(),
                     )
                     return extract_op
+                raise RuntimeError("Unsupported Subscript")
             is_affine = True
             for index in elts:
                 expr = ASTTransformer.build_affine_expr(ctx, index)
--- a/allo/ir/infer.py
+++ b/allo/ir/infer.py
@@ -278,9 +278,9 @@ class TypeInferer(ASTVisitor):
 
     @staticmethod
     def visit_ExtSlice(ctx, node):
-        visit_stmts(ctx, node.dims)
+        stmts = visit_stmts(ctx, node.dims)
         node.shape = tuple()
-        node.dtype = None
+        node.dtype = [stmt.dtype for stmt in stmts]
         return node
 
     @staticmethod
@@ -292,7 +292,7 @@ class TypeInferer(ASTVisitor):
         if node.step is not None:
             visit_stmt(ctx, node.step)
         node.shape = tuple()
-        node.dtype = None
+        node.dtype = (Index(), Index(), Index())
         return node
 
     @staticmethod


allo/ir/builder.py
```diff
@@ -417,14 +417,37 @@ def build_BinOp(ctx, node):
...
+                        indices=target.indices,
+                        ip=ctx.get_ip(),
+                    )
+                    return insert_op
             # Note: Python 3.10 will generate different AST for Subscript compared to Python 3.8
             #       3.10 directly flattens the Index node and removes all the None attributes
             #       inside the node
-            # pylint: disable=redefined-builtin
-            slice = (
+            one_slice = (
```
chhzh123: Probably name it `indices`?

tests/test_tensor.py
```diff
@@ -59,5 +59,69 @@ def outzero_scalar() -> int32:
...
+        return A[1:2, 1:3]
+
+    s = allo.customize(extract, enable_tensor=True)
+    print(s.module)
+
+    mod = s.build()
+    np_A = np.random.randint(0, 10, size=(6, 6)).astype(np.int32)
+    np_Aextract = np_A[1:2, 1:3]
+    np_Aextract_allo = mod(np_A)
+    np.testing.assert_allclose(np_Aextract, np_Aextract_allo, rtol=1e-5)
```
chhzh123: You can directly feed a Numpy array to the kernel now, so you can write

```python
np.testing.assert_allclose(extract(np_A), mod(np_A), rtol=1e-5)
```

tests/test_tensor.py
```diff
@@ -59,5 +59,69 @@ def outzero_scalar() -> int32:
...
+        return A[1, 2]
+
+    s = allo.customize(extract_ele, enable_tensor=True)
+    print(s.module)
+
+    mod = s.build()
+    np_A = np.random.randint(0, 10, size=(6, 6)).astype(np.int32)
+    np_Aextract_ele = np_A[1, 2]
+    np_Aextract_ele_allo = mod(np_A)
+    np.testing.assert_allclose(np_Aextract_ele, np_Aextract_ele_allo, rtol=1e-5)
```
chhzh123: Same. It is not a good name for `np_Aextract_ele`, see [Python naming convention](https://peps.python.org/pep-0008/). We usually use [snake case](https://en.wikipedia.org/wiki/Snake_case) for Python. But to avoid intermediate variable, you can directly pass in the numpy array to the kernel.

tests/test_tensor.py
```diff
@@ -59,5 +59,69 @@ def outzero_scalar() -> int32:
...
+    s = allo.customize(insert, enable_tensor=True)
+    print(s.module)
+
+    mod = s.build()
+    np_A = np.random.randint(0, 10, size=(3, 4, 5)).astype(np.int32)
+    np_B = np.random.randint(0, 10, size=(1, 2, 1)).astype(np.int32)
+    np_Ainsert = np_A.copy()
+    np_Ainsert[1:2, 1:3, 0:1] = np_B
+    np_Ainsert_allo = mod(np_A, np_B)
+    np.testing.assert_allclose(np_Ainsert, np_Ainsert_allo, rtol=1e-5)
```
chhzh123: Same.

tests/test_tensor.py
```diff
@@ -59,5 +59,69 @@ def outzero_scalar() -> int32:
...
+
+
+def test_insert_ele():
+    def insert_ele(A: int32[6, 6], B: int32) -> int32[6, 6]:
```
chhzh123: We normally use big `A`/`B`/`C`/... for matrices, and small `a`/`b`/`c` for scalars

tests/test_tensor.py
```diff
@@ -59,5 +59,69 @@ def outzero_scalar() -> int32:
...
+    s = allo.customize(insert_ele, enable_tensor=True)
+    print(s.module)
+
+    mod = s.build()
+    np_A = np.random.randint(0, 10, size=(6, 6)).astype(np.int32)
+    np_B = np.random.randint(0, 10)
+    np_Ainsert = np_A.copy()
+    np_Ainsert[1, 2] = np_B
+    np_Ainsert_allo = mod(np_A, np_B)
+    np.testing.assert_allclose(np_Ainsert, np_Ainsert_allo, rtol=1e-5)
```
chhzh123: Same

tests/test_tensor.py
```diff
@@ -59,5 +59,69 @@ def outzero_scalar() -> int32:
...
+
+
 if __name__ == "__main__":
-    pytest.main([__file__])
+    # pytest.main([__file__])
+    test_insert()
+    test_insert_ele()
```
chhzh123: Preserve `pytest.main([__file__])`

allo/ir/builder.py
```diff
@@ -569,9 +594,72 @@ def build_Subscript(ctx, node):
         ctx.dim_count = 0
         ctx.affine_vars = []
         index_exprs = []
-        # pylint: disable=redefined-builtin
-        slice = node.slice.value if isinstance(node.slice, ast.Index) else node.slice
-        elts = slice.elts if isinstance(slice, ast.Tuple) else [slice]
+        if ctx.enable_tensor is True and isinstance(node.slice, ast.ExtSlice):
```
chhzh123: If an expression is already a bool type, we don't need `is True`

allo/ir/builder.py
```diff
@@ -569,9 +594,72 @@ def build_Subscript(ctx, node):
...
+                static_sizes=static_sizes,
+                static_strides=static_strides,
+                static_offsets=static_offsets,
+                offsets=[],
+                sizes=[],
+                strides=[],
+                ip=ctx.get_ip(),
+            )
+            return extractslice_op
+        if ctx.enable_tensor is True and isinstance(node.slice, ast.Index):
```
chhzh123: The logic here is a bit weird. Can you put `if ctx.enable_tensor` in front of both branches?

allo/ir/builder.py
```diff
@@ -477,6 +500,8 @@ def build_Assign(ctx, node):
         # Store LHS
         rhs = ASTTransformer.build_cast_op(ctx, rhs, node.value.dtype, node.dtype)
         store_op = ASTTransformer.build_store(ctx, node.targets[0], rhs)
+        if isinstance(store_op, (tensor_d.InsertOp, tensor_d.InsertSliceOp)):
```
chhzh123, line 662: Add a comment:

```python
# Since `tensor_d.InsertOp` returns a copy of the original tensor,
# we need to also update the buffer
```

allo/ir/builder.py
```diff
@@ -417,14 +417,37 @@ def build_BinOp(ctx, node):
     @staticmethod
     def build_store(ctx, node, val):
         if isinstance(node, ast.Subscript):
+            if ctx.enable_tensor is True:
+                node.ctx = ast.Load()
+                target = build_stmt(ctx, node)
```
chhzh123: I didn't get this part. Why do we need to build the node itself here?

allo/ir/builder.py
```diff
@@ -417,14 +417,37 @@ def build_BinOp(ctx, node):
...
+                if isinstance(target, tensor_d.ExtractSliceOp):
```
chhzh123: It's a bit strange to test if the `target` is an `ExtractSliceOp` or an `InsertSliceOp`. What will happen if the target is a node other than those two?

allo/ir/builder.py
mmengjiadai: Do I have to rename `one_slice` in other parts of the code as `indices`? Such as `for one_slice in slices`.

tests/test_tensor.py
mmengjiadai: I saw that in snake case the first letter of each word should be lowercase. Is `np_Aextract_ele` improper because A is uppercase or is it because it is unclear that A and extract is written together?

tests/test_tensor.py
mmengjiadai: I don't know how to input `np_Ainsert` using a python function. Is it ok that I rename it (see question about naming commented above)?

tests/test_tensor.py
chhzh123: It is because you write them together. You can name it `extracted_np_A` for example.

tests/test_tensor.py
chhzh123: It is not the same as `insert(np_A.copy(), np_B)`? Renaming it is also fine

allo/ir/builder.py
chhzh123: It depends on your context. Here is taking multiple indices, so should be named `indices`. For the latter example, you are iterating each slice, so it can be `index`.

allo/backend/llvm.py
```diff
@@ -343,9 +362,11 @@ def __init__(self, mod, top_func_name):
             # Run through lowering passes
             pm = PassManager.parse(
                 "builtin.module(one-shot-bufferize{allow-return-allocs bufferize-function-boundaries},"
+                "expand-strided-metadata,"
                 "func.func(convert-linalg-to-affine-loops),lower-affine)"
             )
             pm.run(self.module.operation)
+            print(self.module)
```
chhzh123: Delete

allo/ir/builder.py
```diff
@@ -515,34 +515,66 @@ def build_BinOp(ctx, node):
...
+                    return insertslice_op
+                if isinstance(node.slice, ast.Index):
+                    index_exprs = ASTTransformer.build_Index(ctx, node)
+                    insert_op = tensor_d.InsertOp(
+                        scalar=val.result,
+                        dest=ctx.buffers[node.value.id].result,
+                        indices=index_exprs,
+                        ip=ctx.get_ip(),
+                    )
+                    return insert_op
```
chhzh123, line 547: Should raise RuntimeError after this line. Otherwise, it is hard to know whether `node.slice` is covered by these two branches.

allo/ir/builder.py
```diff
@@ -711,6 +747,57 @@ def build_affine_expr(ctx, node):
...
+    @staticmethod
+    def build_ExtSlice(ctx, node):
+        # caculate the static offsets, sizes, strides for ExtractSlice and InsertSlice
+        dtype = RankedTensorType(ctx.buffers[node.value.id].result.type).element_type
+        in_shape = RankedTensorType(ctx.buffers[node.value.id].result.type).shape
+        slices = node.slice.dims
+        static_offsets = []
+        static_sizes = []
+        static_strides = []
+        for index, shape in zip(slices, in_shape):
```
chhzh123: Rename `shape` -> `size`

allo/ir/builder.py
```diff
@@ -721,6 +808,34 @@ def build_Subscript(ctx, node):
...
+                    )
+                    return extractslice_op
+                if isinstance(node.slice, ast.Index):
+                    index_exprs = ASTTransformer.build_Index(ctx, node)
+                    extract_op = tensor_d.ExtractOp(
+                        tensor=ctx.buffers[node.value.id].result,
+                        indices=index_exprs,
+                        ip=ctx.get_ip(),
+                    )
+                    return extract_op
```
chhzh123, line 838: Again, raise RuntimeError after this branch.

allo/ir/infer.py
```diff
@@ -276,6 +276,25 @@ def visit_Subscript(ctx, node):
...
+    @staticmethod
+    def visit_Slice(ctx, node):
+        if node.lower is not None:
+            visit_stmt(ctx, node.lower)
+        if node.upper is not None:
+            visit_stmt(ctx, node.upper)
+        if node.step is not None:
+            visit_stmt(ctx, node.step)
+        node.shape = tuple()
+        node.dtype = None
```
chhzh123: I would suggest `node.dtype = (Index(), Index(), Index())`

allo/ir/infer.py
```diff
@@ -276,6 +276,25 @@ def visit_Subscript(ctx, node):
             raise RuntimeError("Can only access bit (slice) for integers")
         return node
 
+    @staticmethod
+    def visit_ExtSlice(ctx, node):
+        visit_stmts(ctx, node.dims)
+        node.shape = tuple()
+        node.dtype = None
```
chhzh123: Similarly, `node.dtype = [stmt.dtype for stmt in stmts]`, where `stmts` is the return of `visit_stmts`

PR #44, opened 2023-08-23
[ir] Use "=" to copy tensor

Crazy-James26:
## Description ##
Add semantics to allow users to use "=" directly create a tensor from another one assigned before or from an input.

### Problems ###
In #31, we didn't the part of tensor. Now we add it.

### Proposed Solutions ###
We import `AllocTensorOp()` from `hcl_mlir.dialects.bufferization` to let users copy a  tensor from another one assigned before or from an input.

### Examples ###
code:
```
def test_copy_arg_tensor():
    M, N = 2, 2

    def kernel(inp: float32[M, N]) -> float32[M, N]:
        A: float32[M, N] = inp
        C: float32[M, N] = A
        return C

    s = allo.customize(kernel, enable_tensor=True)
    print(s.module)

    mod = s.build()
    inp = np.ones((M, N)).astype(np.float32)
    outp = mod(inp)
    np.testing.assert_allclose(inp, outp, rtol=1e-5)
```
MLIR output:
```
module {
  func.func @kernel(%arg0: tensor<2x2xf32>) -> tensor<2x2xf32> attributes {itypes = "_", otypes = "_"} {
    %0 = bufferization.alloc_tensor() copy(%arg0) {name = "A"} : tensor<2x2xf32>
    %1 = bufferization.alloc_tensor() copy(%0) {name = "C"} : tensor<2x2xf32>
    return %1 : tensor<2x2xf32>
  }
}
```
np.testing result: Pass

## Checklist ##

- [x] Code is well-documented

### DIFF
--- a/allo/ir/builder.py
+++ b/allo/ir/builder.py
@@ -483,7 +483,8 @@ class ASTTransformer(ASTBuilder):
             return ASTTransformer.build_library_op(
                 ctx, node=node, op_name=attr, new_args=new_args
             )
-        return opcls[type(node.dtype)](lhs.result, rhs.result, ip=ctx.get_ip())
+        ty_cls = Int if isinstance(node.dtype, Index) else type(node.dtype)
+        return opcls[ty_cls](lhs.result, rhs.result, ip=ctx.get_ip())
 
     @staticmethod
     def build_UnaryOp(ctx, node):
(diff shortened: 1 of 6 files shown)


allo/ir/builder.py
```diff
@@ -722,14 +723,34 @@ def build_AnnAssign(ctx, node):
...
+                    ):
+                        # pylint: disable=unexpected-keyword-arg
+                        alloct_op = buf_d.AllocTensorOp(
+                            tensor_type,
+                            [],
+                            copy=rhs.result,
+                            size_hint=None,
+                            escape=None,
+                            ip=ctx.get_ip(),
+                        )
```
chhzh123: Can you use `tensor_d.EmptyOp` to create a new tensor? According to the [document](https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorempty-tensoremptyop), `EmptyOp` is lowered to `AllocTensorOp` during the bufferization process.

Crazy-James26: Oh, actually I used `tensor_d.EmptyOp` to implement this function at the begining, but it would require me to use  `AllocTensorOp`.  
And I saw that in `build_linalg_op`, Chizhen also used `tensor_d.EmptyOp` to initiate a tensor. But When I open the tensor mode and run related exmples in the tests repo, it would also require us to use `AllocTensorOp` instead of `EmptyOp`.

chhzh123: > but it would require me to use AllocTensorOp.

What do you mean? Are there any errors if you use EmptyOp?

PR #71, opened 2023-09-01
[Frontend] [IR] Extend PyTorch frontend builder and linalg operations

Zhichenzzz:
## Description 
This PR adds new PyTorch frontend builders for Self-Attention Module of transformers and extends linalg library operations to higher dimension by reshaping inputs.

After this, users can directly write PyTorch Transformer Module codes instead of translating them into Allo and call our new API to generate MLIR.

### Solved Problems ###
Following #62, we can capture Self-Attention Module, transfer it to Allo representations, and then lower it to MLIR. The first problem is that in transformer modules, there are more complex matrix operations, like 4D matmul, 3D linear, and reshaping, which are not supported by Linalg Dialects. The other one is how to propagate shapes and translate fx to Allo representations with cleaner codes.

#### IR builder ####

- First, I finished the matrix reshape/view operation to convert it from one type to an equivalent type with a provided shape by using memref/tensor dialect. Now, we can call `allo.view()` to change the tensor shape.

```python
def test1(A: float32[5, 2, 2]) -> float32[5, 4]:
    view = allo.view(A, (5, 4))
    return view
```
- Then, based on this function, we can extend matrix multiplication to higher dimensions like 3D@2D, 4D@4D, etc. Besides, we extend `allo.transpose()` with exchanging two certain axes. These all help us support translating transformer modules.
```python
# 4D matmul
def test2(A: float32[5, 10, 8, 6], B: float32[5, 10, 6, 7]) -> float32[5, 10, 8, 7]:
    C = allo.matmul(A, B)
    return C

def kernel(
        A: float32[M, K, L], B: float32[N, K], C: float32[N]
    ) -> float32[M, L * N]:
        output1 = allo.transpose(A, (-1, -2))
        output2 = allo.linear(output1, B, C)
        output = allo.view(output2, (5, 8))
        return output
```
#### PyTorch frontend ####

- There are some useless shape calculation codes generated like this block. Call `ShapeProp.propagate()` in torch.fx to execute automatic shape propagation of the whole module, and then fold shape calculation operations to release codes. In the following example, `new_shape` calculations will execute internally and Allo directly gets the final shape in `allo.view(x,new_shape)`:
```python
new_shape = x.shape[:-1] + (self.n_heads, -1)
x = x.view(new_shape)
```
- Then, add more call_function builders to support Self-Attention module in `pytorch.py`

### Examples ###
Pass in a PyTorch model using the `allo.frontend.from_pytorch` API, and an LLVM executable will be automatically returned.
```python
[...]

PR #77, opened 2023-09-06
[IR][Pass] Lower linalg.softmax to LLVM using template function

Zhichenzzz:
## Description ##
This PR fixes the [issue](https://reviews.llvm.org/D153422) of linalg library functions not supporting Python for lowering `linalg.SoftmaxOp` to LLVM. Now, #71 `Self_Attention Module` can successfully execute in Allo.

### Problems ###
In LLVM latest version, it cannot support lowering `linalg.SoftmaxOp` to LLVM in Python. This [issue](https://reviews.llvm.org/D153422) needs to be resolved urgently because softmax is a critical operation in transformer models.

### Proposed Solutions ###
Instead of going through the builtin lowering process, we chose to directly invoke the [lowered softmax function](https://github.com/llvm/llvm-project/blob/main/mlir/test/Dialect/Linalg/transform-op-decompose.mlir#L210-L244) and transform it to the users' kernel module. So, first, we should write a complete MLIR code implementation for the softmax function. Then parser it, move the softmax function to the top level, and then call softmax in the top level. After creating new IR operations based on users' `inputs`, use `.replace_all_uses_with` to replace all references entirely. 

### Examples ###
The following example is a part of `Self_Attention Module`, which displays how softmax is called in `transform.py`.
```python
class SelfAttention(nn.Module):
    def __init__(self, hidden_size, n_heads):
    ······

    def permute_for_scores(self, x):
    ······

    def scaled_dot_product(self, q, k, v):
        # (bs, head, seq, hs // head)
        attn_score = torch.matmul(q, k.transpose(-2, -1)) / math.sqrt(
            self.attention_head_size
        )
        # (bs, head, seq, seq)
        attn_probs = F.softmax(attn_score, dim=-1)
        attn_probs = F.dropout(attn_probs, 0.1)
        # (bs, head, seq, hs // head)
        attn = torch.matmul(attn_probs, v)
        return attn

    def forward(self, hidden_states):
    ······
        return output
```
Finally, we transform the previous line to this:
```mlir
%105 = "linalg.softmax"(%99, %102) <{dimension = 0 : i64}> {op_name = "softmax_41"} : (memref<2x12x512x512xf32>, memref<2x12x512x512xf32>) -> memref<2x12x512x512xf32>
loc("-":7:5): error: failed to legalize operation 'linalg.softmax'
loc("-":2:3): error: cannot be converted to LLVM IR: missing `LLVMTranslationDialectInterface` registration for dialect for op: func.func
```

```mlir
%38 = call @softmax_1590757177587462170(%alloc_59, %alloc_61) : (memref<2x12x512x512xf32>, memref<2x12x512x512xf32>) -> memref<2x12x512x512xf32>
```
The above PyTorch module can be lower to MLIR and execute since softmax issue is fixed. The generated softmax IR for this module is shown below:
```mlir
func.func @softmax_1590757177587462170(%arg0: memref<2x12x512x512xf32>, %arg1: memref<2x12x512x512xf32>) -> memref<2x12x512x512xf32> {
    %alloc = memref.alloc() : memref<2x12x512xf32>
    %cst = arith.constant 0xFF800000 : f32
    linalg.fill ins(%cst : f32) outs(%alloc : memref<2x12x512xf32>)
    linalg.generic {indexing_maps = [#map, #map1], iterator_types = ["parallel", "parallel", "parallel", "reduction"]} ins(%arg0 : memref<2x12x512x512xf32>) outs(%alloc : memref<2x12x512xf32>) {
    ^bb0(%in: f32, %out: f32):
      %0 = arith.maxf %in, %out : f32
      linalg.yield %0 : f32
    }
    linalg.generic {indexing_maps = [#map, #map1, #map], iterator_types = ["parallel", "parallel", "parallel", "parallel"]} ins(%arg0, %alloc : memref<2x12x512x512xf32>, memref<2x12x512xf32>) outs(%arg1 : memref<2x12x512x512xf32>) {
    ^bb0(%in: f32, %in_2: f32, %out: f32):
      %0 = arith.subf %in, %in_2 : f32
      %1 = math.exp %0 : f32
      linalg.yield %1 : f32
    }
    %alloc_0 = memref.alloc() : memref<2x12x512xf32>
    %cst_1 = arith.constant 0.000000e+00 : f32
    linalg.fill ins(%cst_1 : f32) outs(%alloc_0 : memref<2x12x512xf32>)
    linalg.generic {indexing_maps = [#map, #map1], iterator_types = ["parallel", "parallel", "parallel", "reduction"]} ins(%arg1 : memref<2x12x512x512xf32>) outs(%alloc_0 : memref<2x12x512xf32>) {
    ^bb0(%in: f32, %out: f32):
      %0 = arith.addf %in, %out : f32
      linalg.yield %0 : f32
    }
    linalg.generic {indexing_maps = [#map, #map1, #map], iterator_types = ["parallel", "parallel", "parallel", "parallel"]} ins(%arg1, %alloc_0 : memref<2x12x512x512xf32>, memref<2x12x512xf32>) outs(%arg1 : memref<2x12x512x512xf32>) {
    ^bb0(%in: f32, %in_2: f32, %out: f32):
      %0 = arith.divf %in, %in_2 : f32
      linalg.yield %0 : f32
    }
    return %arg1 : memref<2x12x512x512xf32>
  }
```
## Still problems ##
Now, since transforming in tensor mode is quite complex, we only support memref mode.

## Checklist ##

- [x] Code is well-documented

### DIFF
--- a/allo/passes.py
+++ b/allo/passes.py
@@ -137,10 +137,6 @@ def generate_input_output_buffers(top_func, flatten=False):
 
 
 def decompose_softmax(module):
-    current_directory = os.path.dirname(os.path.realpath(__file__))
-    file_path = os.path.join(current_directory, "ir/template/softmax_impl.mlir")
-    with open(file_path, "r", encoding="utf-8") as f:
-        softmax_module = f.read()
     with module.context, Location.unknown():
         # get all functions from origin module and find the function to replace
         body_op_to_remove = []
@@ -149,7 +145,7 @@ def decompose_softmax(module):
                 # put softmax function into the module
                 for body_op in op.entry_block.operations:
                     if isinstance(body_op, linalg_d.SoftmaxOp):
-                        generate_softmax(softmax_module, body_op, op)
+                        generate_softmax(body_op, op)
                         body_op_to_remove.append(body_op)
         # need to erase at the end
         for op in body_op_to_remove:
@@ -157,7 +153,11 @@ def decompose_softmax(module):
         return module
 
 
-def generate_softmax(template, target_op, func_op):
+def generate_softmax(target_op, func_op):
+    current_directory = os.path.dirname(os.path.realpath(__file__))
+    file_path = os.path.join(current_directory, "ir/template/softmax_impl.mlir")
+    with open(file_path, "r", encoding="utf-8") as f:
+        template = f.read()
     op_to_remove = []
     # Update arguments of softmax template function
     softmax_mod = Module.parse(template)


allo/ir/softmax_impl.mlir
chhzh123, line 1: Create a new folder `allo/ir/template` and put this piece of code under that folder

allo/customize.py
```diff
@@ -681,7 +681,7 @@ def customize(
     module = ASTTransformer()(ctx, tree)
     if lower_linalg:
         lower_linalg_and_attach_names(module)
-
+    module = softmax_implement(module)
```
chhzh123: It would be better to put this pass under `llvm.py` currently. Also, rename it to `decompose_softmax`

allo/ir/builder.py
```diff
@@ -1436,11 +1436,13 @@ def build_library_op(ctx, node, attr, new_args, dtype=None, shape=None):
                 op = op.owner
             elif attr == "softmax":
                 # TODO: Failed to lower to LLVM, see https://reviews.llvm.org/D153422
+                # Now, we use MLIR transformation implementation to lower softmax
```
chhzh123: `We temporarily replace SoftmaxOp with a predefined lowered function to enable LLVM execution`

allo/ir/softmax_impl.mlir
```diff
@@ -0,0 +1,36 @@
+func.func @softmax(%A: memref<2x16x32xf32>, %B: memref<2x16x32xf32>) -> memref<2x16x32xf32> {
```
chhzh123: Need to add headers like this:
https://github.com/cornell-zhang/hcl-dialect/blob/main/test/Transforms/compute/cascade.mlir#L1-L2

chhzh123: Also, provide the original link of this test file (that one included in the LLVM repository)

allo/ir/transform.py
```diff
@@ -242,3 +250,99 @@ def find_func_in_module(module, func_name):
         if isinstance(op, func_d.FuncOp) and op.name.value == func_name:
             return op
     return None
+
+
+def softmax_implement(module):
```
chhzh123: `decompose_softmax`. Put it under `passes.py`

tests/test_linalg.py
```diff
@@ -236,15 +236,20 @@ def kernel(A: float32[M, K], B: float32[K, M]) -> float32[M, M]:
 
 def test_linalg_softmax():
     # TODO: failed to lower to LLVM, see https://reviews.llvm.org/D153422
-    M = 10
-    K = 15
+    # Now, we use MLIR transformation implementation to lower softmax
```
chhzh123: See the above comments

allo/ir/transform.py
```diff
@@ -242,3 +250,99 @@ def find_func_in_module(module, func_name):
...
+                                    == "[#linalg.iterator_type<parallel>, #linalg.iterator_type<parallel>, #linalg.iterator_type<parallel>]"
+                                ):
+                                    softmax_op.attributes[
+                                        "indexing_maps"
+                                    ] = ArrayAttr.get(
+                                        [affine_map_0, affine_map_1, affine_map_0]
+                                    )
+                                    softmax_op.attributes[
+                                        "iterator_types"
+                                    ] = ArrayAttr.get(iter_types_1)
```
chhzh123: It would be better to move this part to a new function (e.g., `generate_softmax`) to avoid excessive indentation.

tests/test_linalg.py
```diff
@@ -236,15 +236,20 @@ def kernel(A: float32[M, K], B: float32[K, M]) -> float32[M, M]:
...
+    M = 8
+    K = 16
+    N = 32
 
-    def kernel(A: float32[M, K]) -> float32[M, K]:
+    def kernel(A: float32[M, K, N]) -> float32[M, K, N]:
         outs = allo.softmax(A)
```
chhzh123: As you create separate softmax functions for each function call, can you invoke softmax function multiple times in this case to see whether it works properly?

allo/ir/transform.py
```diff
@@ -242,3 +250,99 @@ def find_func_in_module(module, func_name):
...
+                                alloc_op = memref_d.AllocOp(
+                                    MemRefType.get(
+                                        shape[:-1], MemRefType(in_types[0]).element_type
+                                    ),
+                                    [],
+                                    [],
+                                    ip=InsertionPoint(softmax_op),
+                                )
+                                softmax_op.result.replace_all_uses_with(alloc_op.result)
+                                softmax_op.operation.erase()
```
chhzh123: I would suggest using a list to store the operations that need to be removed and remove them together at the end of the pass, as shown in the following example:
https://github.com/cornell-zhang/allo/blob/573cda743b754aeb7623f1507222535d844f4bb0/allo/customize.py#L546-L548

Otherwise, it is error-prone to erase an operation at the same time traverse the operation list.

allo/ir/transform.py
```diff
@@ -242,3 +250,99 @@ def find_func_in_module(module, func_name):
...
+                                var_str_0 = ", ".join(
+                                    [f"d{i}" for i in range(len(shape))]
+                                )
+                                var_str_1 = ", ".join(
+                                    [f"d{i}" for i in range(len(shape) - 1)]
+                                )
+                                affine_map_0 = AffineMapAttr.parse(
+                                    f"affine_map<({var_str_0})->({var_str_0})>"
+                                )
+                                affine_map_1 = AffineMapAttr.parse(
```
chhzh123: Can you give more meaningful names or add more comments? What are these two affine maps?

allo/ir/transform.py
```diff
@@ -242,3 +250,99 @@ def find_func_in_module(module, func_name):
...
+                                ] * (len(shape) - 1) + [
+                                    Attribute.parse("#linalg.iterator_type<reduction>")
+                                ]
+                                iter_types_1 = [
+                                    Attribute.parse("#linalg.iterator_type<parallel>")
+                                ] * len(shape)
+                                if (
+                                    str(softmax_op.attributes["iterator_types"])
+                                    == "[#linalg.iterator_type<parallel>, #linalg.iterator_type<parallel>, #linalg.iterator_type<reduction>]"
+                                ):
```
chhzh123: Can you explain what are these two branches used for?

chhzh123: Also, you need to raise an error for the `else` branch

allo/ir/transform.py
```diff
@@ -242,3 +250,99 @@ def find_func_in_module(module, func_name):
...
+                        softmax_func.move_before(op)
+                        func_d.CallOp(
+                            [body_op.output.type],
+                            FlatSymbolRefAttr.get(f"softmax_{hash(body_op)}"),
+                            [body_op.input, body_op.output],
+                            ip=InsertionPoint(body_op),
+                        )
+                        shape = MemRefType(in_types[0]).shape
+
+                        for softmax_op in softmax_func.entry_block.operations:
```
chhzh123: Add a comment before this line:
```
Update memref shapes and dtypes in the softmax function
```

allo/backend/llvm.py
```diff
@@ -277,6 +277,8 @@ def __init__(self, mod, top_func_name, ext_libs=None):
         # Copy the module to avoid modifying the original one
         with Context() as ctx:
             hcl_d.register_dialect(ctx)
+            # Decompose softmax
```
chhzh123: Delete comment

allo/backend/llvm.py
```diff
@@ -277,6 +277,8 @@ def __init__(self, mod, top_func_name, ext_libs=None):
...
+            mod = decompose_softmax(mod)
```
chhzh123: Should take in `self.module` and place before `_mlir_lower_pipeline`
https://github.com/cornell-zhang/allo/blob/a5215b809f0e48682759a00f71df21299415beb0/allo/backend/llvm.py#L288

examples/self_attention.py
```diff
@@ -31,7 +31,7 @@ def scaled_dot_product(self, q, k, v):
         )
         # (bs, head, seq, seq)
         attn_probs = F.softmax(attn_score, dim=-1)
-        attn_probs = F.dropout(attn_probs, 0.1)
+        attn_probs = F.dropout(attn_probs, 0)
```
chhzh123: Can't dropout work for `p=0.1`? It should act like an identity function when you set up the `.eval()` mode.
https://leimao.github.io/blog/Dropout-Explained/

allo/passes.py
```diff
@@ -122,3 +134,100 @@ def generate_input_output_buffers(top_func, flatten=False):
...
+                    if isinstance(body_op, linalg_d.SoftmaxOp):
+                        generate_softmax(softmax_module, body_op, op)
+                        body_op_to_remove.append(body_op)
+        # need to erase at the end
+        for op in body_op_to_remove:
+            op.operation.erase()
+        return module
+
+
+def generate_softmax(module, body_op, init_op):
```
chhzh123: It would be better to call it `template` instead of `module`

allo/passes.py
```diff
@@ -122,3 +134,100 @@ def generate_input_output_buffers(top_func, flatten=False):
...
+    op_to_remove = []
+    # get softmax function
+    softmax_mod = Module.parse(module)
+    softmax_func = softmax_mod.body.operations[0]
+    softmax_func.attributes["sym_name"] = StringAttr.get(f"softmax_{hash(body_op)}")
+    args = softmax_func.arguments
```
chhzh123, line 166: Add comment: `Update arguments of softmax template function`

allo/passes.py
```diff
@@ -122,3 +134,100 @@ def generate_input_output_buffers(top_func, flatten=False):
...
+    softmax_func.move_before(init_op)
+    func_d.CallOp(
+        [body_op.output.type],
+        FlatSymbolRefAttr.get(f"softmax_{hash(body_op)}"),
+        [body_op.input, body_op.output],
+        ip=InsertionPoint(body_op),
+    )
+    # Update memref shapes and dtypes in the softmax function
+    shape = MemRefType(in_types[0]).shape
+    for softmax_op in softmax_func.entry_block.operations:
```
chhzh123: Not `softmax_op` but general `op`:

```python
for op in softmax_func.entry_block.operations:
```

allo/passes.py
chhzh123: `body_op` -> `target_op`, `init_op` -> `func_op`

tests/test_linalg.py
```diff
@@ -236,15 +236,22 @@ def kernel(A: float32[M, K], B: float32[K, M]) -> float32[M, M]:
...
+    K = 16
+    N = 32
 
-    def kernel(A: float32[M, K]) -> float32[M, K]:
-        outs = allo.softmax(A)
-        return outs
+    def kernel(A: float32[M, K, N], B: float32[K, N]) -> float32[M, K, N]:
+        C = allo.softmax(A)
+        D = allo.softmax(B)
+        return C + D
```
chhzh123: Can you also test multiple softmax in nested functions? For example, `kernel` and `foo` both have softmax functions, and `kernel` calls `foo` in its body.

examples/self_attention.py
Zhichenzzz: Yes, when set $p=0.1$, the output results will exceed the error range.

Zhichenzzz:
```
Mismatched elements: 596711 / 786432 (75.9%)
Greatest absolute difference: 0.04492151737213135 at index (0, 325, 759) (up to 0.001 allowed)
Greatest relative difference: 2841.284842796809 at index (1, 155, 223) (up to 0.001 allowed)
```

chhzh123: It does not make sense for dropout to generate different results. Can you try generating PyTorch results before creating an fx module?
https://github.com/cornell-zhang/allo/blob/81b7c70912c959ca7d2fd252a48d76ccbbaeccb4/examples/self_attention.py#L64

chhzh123: Also, try some small examples that contain a dropout function and see its behavior during inference. It should act like an identity function.

Zhichenzzz: Yeah, it should act like identity according to the comment:
```html
        This has any effect only on certain modules. See documentations of
        particular modules for details of their behaviors in training/evaluation
        mode, if they are affected, e.g. :class:`Dropout`, :class:`BatchNorm`,
        etc.

        This is equivalent with :meth:`self.train(False) <torch.nn.Module.train>`.
```

Zhichenzzz: > It does not make sense for dropout to generate different results. Can you try generating PyTorch results before creating an fx module?
> 
> https://github.com/cornell-zhang/allo/blob/81b7c70912c959ca7d2fd252a48d76ccbbaeccb4/examples/self_attention.py#L64

This can not work as well. I will test some small test cases.
```python
model = SelfAttention(hidden_size, n_heads).eval()
example_inputs = [torch.rand(batch_size, seq_len, hidden_size)]
golden = model(*example_inputs)
llvm_mod = allo.frontend.from_pytorch(
    model, example_inputs=example_inputs, verbose=True
)
```

Zhichenzzz: I see. `.eval()` only works when calling class:`Dropout`, while we use dropout by `attn_probs = F.dropout(attn_probs, 0.1)`. So, we should move this line to `def __init__`.

chhzh123: I see. You can do that

allo/passes.py
```diff
@@ -122,3 +134,100 @@ def generate_input_output_buffers(top_func, flatten=False):
                 break
         func_type = FunctionType.get(new_in_types, new_out_types)
         top_func.attributes["function_type"] = TypeAttr.get(func_type)
+
+
+def decompose_softmax(module):
+    current_directory = os.path.dirname(os.path.realpath(__file__))
+    file_path = os.path.join(current_directory, "ir/template/softmax_impl.mlir")
+    with open(file_path, "r", encoding="utf-8") as f:
+        softmax_module = f.read()
```
chhzh123: Only when we find `linalg.softmax`, we read the template function, so this part should be put under `generate_softmax`

allo/passes.py
```diff
@@ -122,3 +134,100 @@ def generate_input_output_buffers(top_func, flatten=False):
...
+                    if isinstance(body_op, linalg_d.SoftmaxOp):
+                        generate_softmax(softmax_module, body_op, op)
+                        body_op_to_remove.append(body_op)
+        # need to erase at the end
+        for op in body_op_to_remove:
+            op.operation.erase()
+        return module
+
+
+def generate_softmax(template, target_op, func_op):
```
chhzh123: Then you can remove the `template` argument here

PR #79, opened 2023-09-08
[IR]Cast tensors with integer types

mmengjiadai:
## Description ##
This PR fix the problem of not being able to broadcast tensors with int types when doing binary ops.

### Problems ###
We implement bit extensions when doing binary operation with integers. Previously we have already achieved casting result tensors to a new type. However MLIR requires casting for input tensors as well, which causes bugs when doing binops with tensors.

### Proposed Solutions ###
By adding argument `shape` to `build_cast_op()`, we use the inputs to identify cases with both integers and tensors. In such cases, we use `linalg.generic` to cast tensors by element. 
```
cast_op = linalg_d.GenericOp(
                    indexing_maps=indexing_maps_attr,
                    ip=ctx.get_ip(),
                    inputs=[op.result],
                    outputs=[tensor_op.result],
                    result_tensors=[RankedTensorType.get(shape, res_type.build())],
                    iterator_types=iterator_types_attr,
                )
``` 
Notice that `tensor_op` in outputs is a new `tensor.empty ` operation constructed with result type and shape.
We then add a block to `linalg.generic` for casting and yielding.
```
block_arg_types = [src_type.build(), res_type.build()]
                block = cast_op.regions[0].blocks.append(*block_arg_types)
                ctx.set_ip(block)
                # add cast op to block
                yield_value = opcls(
                    res_type.build(), block.arguments[0], ip=ctx.get_ip()
                )
                linalg_d.YieldOp([yield_value], ip=ctx.get_ip())
                ctx.pop_ip()
```
`opcls` is a `arith op` determined by the inputs `src_type` and `res_type`. By using `block.arguments` as its argument, `opcls` will cast `inputs` of linalg.generic elementwise.

### Examples ###
Testcase with the following kernel can successfully `customize` and `build`.
```
def kernel() -> int32[2, 2]:
        A: int32[2, 2] = 1
        B = A + 1
        C = B - 3
        return C
```
mlir generated by a typical case of casting int tensors is as follows. 
```
 %12 = tensor.empty() : tensor<2x2xi34>
    %13 = linalg.generic {indexing_maps = [#map, #map], iterator_types = ["parallel", "parallel"]} ins(%11 : tensor<2x2xi33>) outs(%12 : tensor<2x2xi34>) {
    ^bb0(%in: i33, %out: i34):
      %24 = arith.extsi %in : i33 to i34
      linalg.yield %24 : i34
    } -> tensor<2x2xi34>
```

## Checklist ##

- [x] Code is well-documented

### DIFF
--- a/tests/test_tensor.py
+++ b/tests/test_tensor.py
@@ -3,10 +3,8 @@
 
 import pytest
 import allo
-from allo.ir.types import int32, float32
+from allo.ir.types import int32, float32, Int, UInt
 import numpy as np
-from allo.ir.types import Int, UInt, Float, Fixed, UFixed, uint1, int32, float32, index
-import allo.ir.types as T
 
 
 def test_same():


tests/test_tensor.py
```diff
@@ -251,5 +251,19 @@ def kernel() -> float32[M, K]:
     np.testing.assert_allclose(f(), kernel(), atol=1e-4)
 
 
+def test_broadcast_int():
+    def kernel() -> int32[2, 2]:
+        A: int32[2, 2] = 1
```
chhzh123: Can you explicitly pass in a random NumPy array? Also, a 2x2 shape is too small for testing

tests/test_tensor.py
```diff
@@ -251,5 +251,19 @@ def kernel() -> float32[M, K]:
...
+    s = allo.customize(kernel, enable_tensor=True)
+    print(s.module)
+    f = s.build()
+    np.testing.assert_allclose(f(), kernel(), atol=1e-4)
+
+
 if __name__ == "__main__":
-    pytest.main([__file__])
+    # pytest.main([__file__])
+    test_broadcast_int()
```
chhzh123: Recover `pytest.main([__file__])`

tests/test_tensor.py
```diff
@@ -5,6 +5,8 @@
 import allo
 from allo.ir.types import int32, float32
 import numpy as np
+from allo.ir.types import Int, UInt, Float, Fixed, UFixed, uint1, int32, float32, index
+import allo.ir.types as T
```
chhzh123: Where `T` is used? Avoid importing unused names, and merge L6 and L8.

PR #83, opened 2023-09-13
[PyTorch] Support lowering BERT model from HuggingFace

Zhichenzzz:
## Description ##
Based on #80, this PR support connecting [Huggingface BertLayer](https://huggingface.co/docs/transformers/model_doc/bert#transformers.BertModel) to Allo framework, and also fix some problems in translation process.

### Proposed Solutions ###
Basically, since we finished `LayerNorm` and `Gelu` operations, users can write Bert Module in PyTorch and capture it by `torch.py` to traverse the nodes to different Allo library functions. 

Besides, we updated the frontend to support capturing the modules from Transformers. Later on, we will support more modules in [Huggingface Transformer](https://huggingface.co/docs/transformers/index#transformers).

In `fx.trace()`, users should partially specialize the inputs to `concrete_args` because of `if-else` branches in Huggingface inner modules. And we capture these specialized inputs into `ShapeProp(gm).propagate` to generate shapes passes.

### Examples ###
Users can directly capture modules from Huggingface and lower it to MLIR. Here is the example:
```python
from transformers import AutoConfig, BertLayer

config = AutoConfig.from_pretrained("bert-base-uncased")
model = BertLayer(config).eval()
example_inputs = [torch.rand(batch_size, seq_len, hidden_size)]
llvm_mod = allo.frontend.from_pytorch(
        module,
        example_inputs=example_inputs,
        verbose=False,
    )
```
The generated MLIR (part) is shown below:
```mlir
module {
  func.func private @layernorm_8738864313914(memref<2x512x768xf32>, memref<768xf32>, memref<768xf32>) -> memref<2x512x768xf32>
  func.func private @gelu_8738864313965(memref<2x512x3072xf32>) -> memref<2x512x3072xf32>
  func.func private @layernorm_8738864314037(memref<2x512x768xf32>, memref<768xf32>, memref<768xf32>) -> memref<2x512x768xf32>
  func.func @forward(%arg0: memref<2x512x768xf32>) -> memref<2x512x768xf32> attributes {itypes = "_", otypes = "_"} {
   ...
    %alloc = memref.alloc() : memref<2x512x768xf32>
    %c0_i32 = arith.constant 0 : i32
    %16 = arith.sitofp %c0_i32 : i32 to f32
    linalg.fill {op_name = "linear_init_zero_0"} ins(%16 : f32) outs(%alloc : memref<2x512x768xf32>)
    %alloc_0 = memref.alloc() : memref<768x768xf32>
[...]

PR #85, opened 2023-09-18
[PyTorch] Add customized fx.tracer and lower GPT2 model

Zhichenzzz:
## Description ##
This PR extends #83 to support lowering [GPT2](https://huggingface.co/gpt2) in PyTorch to MLIR.
### Problems ###
Last PR, we finished lowering [BERT](https://huggingface.co/bert-base-uncased) from Huggingface to MLIR. To expand the application scope of Allo's transformers, we have added support for the GPT2 model.

### Proposed Solutions ###
Basically, in order to support mask operations in GPT2, we add `allo.ones` and `allo.tril` to generate mask tensors. Then, we try to trace PyTorch Module to graph module. 

In this progress, `torch.fx.symbolic_trace` fails on `torch.ones` because tensor constructors are currently not traceable. We solved this according to [this issue](https://github.com/pytorch/pytorch/issues/51803) and built a customized tracer to override these function. Basically, we patched the torch functions which can not be traced ,then call `fx.trace` to trace the decorated module, after that, recovering the origin function.
```python
def trace(self):
    for name, (wrapper, orig) in self.patched_torch_methods.items():
        setattr(torch, name, wrapper)
        self.orig_fns.add(orig)
    try:
        graph = super().trace(self.model, self.concrete_args)
    finally:
        for name, (_, orig) in self.patched_torch_methods.items():
            setattr(torch, name, orig)
    return graph
```
### Examples ###
Firstly, the usage of `allo.ones()` and `allo.tril()` is displayed below, which can play a similar function as `numpy.ones/tril`.
```python
def test_triangular_matrix():
    def foo(A: float32[M, N]) -> float32[M, N]:
        outp = allo.tril(A)
        return outp
    def kernel(A: float32[M, N], B: float32[M, N]) -> float32[M, M]:
        C = allo.ones((10, 15), dtype=float)
        D = foo(B) + C
        output = allo.matmul(A, D.T)
        return output
    s = allo.customize(kernel)
    mod = s.build()
    outp = mod(A, B)
    np_outp = kernel(A, B)
    np.testing.assert_allclose(outp, np_outp, atol=1e-5)
```
Based on the above, we can lower the GPT2 module into MLIR:
```python
class GPT2(nn.Module):
    def __init__(self, vocab_size, n_embd, n_head, n_layers):
        super(GPT2, self).__init__()
        self.transformer_blocks = nn.ModuleList(
            [TransformerBlock(n_embd, n_head, n_embd * 4) for _ in range(n_layers)]
        )
        self.ln_f = nn.LayerNorm(n_embd)
        self.fc = nn.Linear(n_embd, vocab_size)
    def forward(self, x):
        for block in self.transformer_blocks:
            x = block(x)
        x = self.ln_f(x)
        x = self.fc(x)
        return x
class TransformerBlock(nn.Module):
...
```

### Further Step ###
We will try to lower the GPT2 from Huggingface directly and test its correction.
## Checklist ##

- [x] Code is well-documented

### DIFF
--- a/allo/frontend/pytorch.py
+++ b/allo/frontend/pytorch.py
@@ -12,13 +12,13 @@ try:
     from torch.nn import functional as F
     from torch.fx.graph_module import GraphModule
     from torch.fx.passes.shape_prop import ShapeProp, TensorMetadata
+    from .tracer import AlloTracer
 except ImportError:
     pass
 
 from .. import dsl
 from ..ir import types
 from ..customize import customize
-from .tracer import AlloTracer
 
 
 def from_pytorch(model, example_inputs, verbose=False):


allo/dsl.py
```diff
@@ -145,3 +145,11 @@ def layernorm(x, gamma, beta, eps: float = 1e-5):
 
 def gelu(x):
     return 0.5 * x * (1 + np.tanh(np.sqrt(2 / np.pi) * (x + 0.044715 * x**3)))
+
+
+def ones(shape, dtype=None, order="C"):
+    return np.ones(shape, dtype=dtype, order=order)
```
chhzh123: No need to add `order` argument

allo/frontend/pytorch.py
```diff
@@ -9,13 +9,15 @@
...
-    from torch.fx.passes.shape_prop import ShapeProp
+    from torch.fx.graph_module import GraphModule
+    from torch.fx.passes.shape_prop import ShapeProp, TensorMetadata
 except ImportError:
     pass
 
 from .. import dsl
 from ..ir import types
 from ..customize import customize
+from .tracer import CustomedTracer
```
chhzh123: Rename `AlloTracer`

allo/frontend/pytorch.py
```diff
@@ -148,22 +159,41 @@ def build_call_method(self, node):
         )
 
     def build_output(self, node):
-        name = get_var_name(node.args[0])
         # Currently we only support return a single value
```
chhzh123: It would be better to raise a warning if the model returns multiple values. You can use `warnings.warn` to warn users that we only return the first output

allo/frontend/tracer.py
```diff
@@ -0,0 +1,44 @@
...
+# SPDX-License-Identifier: Apache-2.0
+# https://github.com/huggingface/transformers/blob/main/src/transformers/utils/fx.py
+
+import torch
+from torch.fx import Tracer
+from transformers.utils.fx import _gen_constructor_wrapper
+
+
+# https://github.com/pytorch/pytorch/issues/51803
+class CustomedTracer(Tracer):
```
chhzh123: `AlloTracer`

allo/frontend/tracer.py
```diff
@@ -0,0 +1,44 @@
+# Copyright Allo authors. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+# https://github.com/huggingface/transformers/blob/main/src/transformers/utils/fx.py
+
+import torch
+from torch.fx import Tracer
+from transformers.utils.fx import _gen_constructor_wrapper
```
chhzh123: I think you can directly copy these few lines to Allo and add a reference to HuggingFace. It seems this function does not have other dependencies.
https://github.com/huggingface/transformers/blob/main/src/transformers/utils/fx.py#L687-L705

allo/frontend/pytorch.py
```diff
@@ -4,18 +4,21 @@
...
-    from torch.fx.passes.shape_prop import ShapeProp
+    from torch.fx.graph_module import GraphModule
+    from torch.fx.passes.shape_prop import ShapeProp, TensorMetadata
 except ImportError:
     pass
 
 from .. import dsl
 from ..ir import types
 from ..customize import customize
+from .tracer import AlloTracer
```
chhzh123: Also need to put this line into the try block

PR #86, opened 2023-09-26
[Backend][LLVM] Add support for returning multiple numpy arrays

zzzDavid:
## Description ##
This PR adds support for returning multiple numpy arrays from the MLIR execution engine. 

### Problems ###
Previously we are limited by MLIR execution engine's interface: we can only pass in one return pointer to the `execution_engine.invoke()` function, so we supported only for one return value. This PR packs multiple return pointers to a struct, and pass the pointer to that struct to the execution engine. 

For example, if we have two return memrefs, this is what's happening under the hood:
```c++
struct Result {
  mlir::OwningMemRef<float, 4> first;
  mlir::OwningMemRef<float, 4> second;
};
Result result;
engine->invoke("name", &result);
```
Note that LLVM does not support multi-result function, so there are two conversions happening here:
1. Multiple results get packed into a struct,
2. Then the single-struct result is transformed into a trailing pointer argument because we requested a C-friendly interface by `llvm.emit_c_interface`.

### Proposed Solutions ###
To support more than one return tensors, this PR makes changes in the following components:
- `TypeInferer`: allow `FunctionDef` node to have multiple return dtypes and shapes.
- `ASTTransformer`, which is the IR builder: added a case for return ops with multiple operands.
- `LLVMModule`: added a case to pack multiple return memref descriptors to one struct and then unpack them after invoking the execution engine.

There is one assumption about the return values after this change:
🔔  When there are more than one return values, all of the values are tensors instead of scalars.
This assumption is checked in the llvm backend.

Two new test cases are added:
- `tests/test_builder.py::test_multiple_returns_1D`
- `tests/test_builder.py::test_multiple_returns_4D`

### Examples ###
With this PR, now people can write more than one return values, and the semantics for calling the kernel is just like a python function call:

```python
def test_multiple_returns_4D():
    M = 10

    def kernel(
        A: float32[M, M, M, M], B: float32[M, M, M, M]
    ) -> (float32[M, M, M, M], float32[M, M, M, M]):
        res0: float32[M, M, M, M] = 0
        res1: float32[M, M, M, M] = 0
        for i in range(M):
            for j in range(M):
                for k in range(M):
                    for l in range(M):
                        res0[i, j, k, l] = A[i, j, k, l] + 1
                        res1[i, j, k, l] = B[i, j, k, l] + 1
        return res0, res1

    s = allo.customize(kernel)
    mod = s.build()
    np_A = np.random.random((M, M, M, M)).astype(np.float32)
    np_B = np.random.random((M, M, M, M)).astype(np.float32)
    np_res0, np_res1 = mod(np_A, np_B)
    np.testing.assert_allclose(np_res0, np_A + 1)
    np.testing.assert_allclose(np_res1, np_B + 1)
```

## Checklist ##

- [x] Code is well-documented

### DIFF
--- a/allo/backend/llvm.py
+++ b/allo/backend/llvm.py
@@ -17,8 +17,6 @@ from hcl_mlir.execution_engine import ExecutionEngine
 from hcl_mlir.runtime import (
     get_ranked_memref_descriptor,
     make_nd_memref_descriptor,
-    # ranked_memref_to_numpy,
-    to_numpy,
 )
 from hcl_mlir.exceptions import DTypeWarning
 from ..ir.transform import find_func_in_module
@@ -42,25 +40,8 @@ from ..utils import (
     get_np_struct_type,
     create_output_struct,
     extract_out_np_arrays_from_out_struct,
+    ranked_memref_to_numpy,
 )
-
-
-def ranked_memref_to_numpy(ranked_memref):
-    """Converts ranked memrefs to numpy arrays."""
-    # A temporary workaround for issue
-    # https://discourse.llvm.org/t/setting-memref-elements-in-python-callback/72759
-    contentPtr = ctypes.cast(
-        ctypes.addressof(ranked_memref[0].aligned.contents)
-        + ranked_memref[0].offset * ctypes.sizeof(ranked_memref[0].aligned.contents),
-        type(ranked_memref[0].aligned),
-    )
-    np_arr = np.ctypeslib.as_array(contentPtr, shape=ranked_memref[0].shape)
-    strided_arr = np.lib.stride_tricks.as_strided(
-        np_arr,
-        np.ctypeslib.as_array(ranked_memref[0].shape),
-        np.ctypeslib.as_array(ranked_memref[0].strides) * np_arr.itemsize,
-    )
-    return to_numpy(strided_arr)
 
 
 def invoke_mlir_parser(mod: str):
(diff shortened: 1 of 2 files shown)


allo/backend/llvm.py
```diff
@@ -238,60 +238,136 @@ def __call__(self, *args):
...
+            # Create a struct
+            out_struct = create_output_struct(out_memref_descs)
+            return_ptr = ctypes.pointer(ctypes.pointer(out_struct))
 
         # 3. Invoke the function and return the result
-        if len(shape) > 0:
+        if len(result_types) == 1:
+            result_type, shape = result_types[0]
+            if len(shape) > 0:  # single return, memref
+                # INVOKE
```
chhzh123, line 301: What's the `INVOKE` comment here?

allo/utils.py
```diff
@@ -332,3 +332,66 @@ def cast_func(x):
...
+
+def ranked_memref_descriptor_to_numpy(ranked_memref):
+    """Converts ranked memrefs to numpy arrays."""
+    np_arr = np.ctypeslib.as_array(ranked_memref.aligned, shape=ranked_memref.shape)
+    strided_arr = np.lib.stride_tricks.as_strided(
+        np_arr,
+        np.ctypeslib.as_array(ranked_memref.shape),
+        np.ctypeslib.as_array(ranked_memref.strides) * np_arr.itemsize,
+    )
+    return to_numpy(strided_arr)
```
chhzh123: Why not directly import from `hcl_mlir.runtime`?

chhzh123: Also, we have the latest version of `ranked_memref_to_numpy` that fixes an MLIR bug. See:
https://github.com/cornell-zhang/allo/blob/main/allo/backend/llvm.py#L46-L61

allo/backend/llvm.py
zzzDavid: Ah I added a `INVOKE` comment above every `self.execution_engine.invoke` call, it's only a visual guide

allo/utils.py
zzzDavid: You are absolutely right, I missed the `ranked_memref_to_numpy` in `llvm.py`. Now I have merged mine with this function. The reason I needed a special one is that MLIR's `ranked_memref_to_numpy` accepts a "pointer to memref descriptor", while I only have memref descriptors after I unpack them from the output struct.

PR #87, opened 2023-09-28
[IR] Add support for concatenating two tensors and linear without bias

Zhichenzzz:
## Description ##
In order to support the following [kv cache](https://www.dipkumar.dev/becoming-the-unbeatable/posts/gpt-kvcache/) in GPT module optimizations, this PR focus on supporting concatenating two tensors and updating the function of allo.linear.

### Proposed Solutions ###
We proposed `allo.concat` to support the concatenation of two matrixes. However, there are some difference between tensor dialect and memref diaclect since [tensor.insertslice](https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorinsert_slice-tensorinsertsliceop) can not be implemented in the latter. So we divided the op into two kinds of realizations.

Besides, we fixed the previous issue in not supporting `bias=False` in linear operation.

### Examples ###
Here is the example to illustrate the above proposal.
```python
def test_concatenate(enable_tensor):
    def linear_no_bias(A: int32[20, 10], B: int32[6, 10]) -> int32[20, 6]:
        return allo.linear(A, B)

    def concat(
        A: int32[3, 2, 15, 6], B: int32[3, 2, 5, 6], C: int32[20, 10], D: int32[6, 10]
    ) -> int32[3, 2, 20, 6]:
        out1 = allo.concat(A, B, axis=2)
        output = linear_no_bias(C, D) + out1
        return output
```
The concatenations involved are below：

```mlir
%inserted_slice = tensor.insert_slice %arg0 into %0[0, 0, 0, 0] [3, 2, 15, 6] [1, 1, 1, 1] : tensor<3x2x15x6xi32> into tensor<3x2x20x6xi32>
%inserted_slice_0 = tensor.insert_slice %arg1 into %inserted_slice[0, 0, 15, 0] [3, 2, 5, 6] [1, 1, 1, 1] {name = "out1"} : tensor<3x2x5x6xi32> into tensor<3x2x20x6xi32>
```

```mlir
%subview = memref.subview %alloc[0, 0, 0, 0] [3, 2, 15, 6] [1, 1, 1, 1] : memref<3x2x20x6xi32> to memref<3x2x15x6xi32, strided<[240, 120, 6, 1]>>
memref.copy %arg0, %subview : memref<3x2x15x6xi32> to memref<3x2x15x6xi32, strided<[240, 120, 6, 1]>>
%subview_0 = memref.subview %alloc[0, 0, 15, 0] [3, 2, 5, 6] [1, 1, 1, 1] : memref<3x2x20x6xi32> to memref<3x2x5x6xi32, strided<[240, 120, 6, 1], offset: 90>>
memref.copy %arg1, %subview_0 : memref<3x2x5x6xi32> to memref<3x2x5x6xi32, strided<[240, 120, 6, 1], offset: 90>>
```

Later, we will try to finish kv cache optimization in GPT module by calling these functions.
## Checklist ##

- [x] Code is well-documented

### DIFF
--- a/allo/ir/builder.py
+++ b/allo/ir/builder.py
@@ -29,8 +29,8 @@ from hcl_mlir.ir import (
     TypeAttr,
     ArrayAttr,
     Attribute,
-    Type,
 )
+from hcl_mlir.ir import Type as MLIRType
 from hcl_mlir.dialects import (
     hcl as hcl_d,
     func as func_d,
@@ -1599,10 +1599,10 @@ class ASTTransformer(ASTBuilder):
                 for size in concat_shape[0][axis:]:
                     memref_offsets *= size
                 result = [
-                    Type.parse(
+                    MLIRType.parse(
                         f"memref<{'x'.join([str(x) for x in concat_shape[0]])}x{dtype}, strided<{memref_strides}>>"
                     ),
-                    Type.parse(
+                    MLIRType.parse(
                         f"memref<{'x'.join([str(x) for x in concat_shape[1]])}x{dtype}, strided<{memref_strides}, offset: {memref_offsets}>>"
                     ),
                 ]


allo/ir/builder.py
```diff
@@ -1554,6 +1555,90 @@ def build_library_op(ctx, node, attr, new_args, dtype=None, shape=None):
...
+                    offsets=[],
+                    sizes=[],
+                    strides=[],
+                    ip=ctx.get_ip(),
+                )
+                memref_d.CopyOp(
+                    new_args[0].result,
+                    op_,
+                    ip=ctx.get_ip(),
+                )
```
chhzh123, line 1624: Why do we need a memcopy here?

allo/ir/builder.py
```diff
@@ -1714,12 +1799,18 @@ def build_library_op(ctx, node, attr, new_args, dtype=None, shape=None):
...
                     ctx, node, inner_attr, [new_args[0], A_T]
                 )
-                dims = list(range(len(node.shape) - 1))
-                bias = ASTTransformer.build_broadcast_op(
-                    ctx, new_args[2], node.dtype, node.shape[-1:], node.shape, dims
-                )
-                add = ASTTransformer.build_library_op(ctx, node, "add", [matmul, bias])
-                return add
+                # bias = True
+                if len(new_args) == 3:
```
chhzh123: We probably need to have a path to handle `kwargs`. Otherwise, when users call `allo.linear(X, A, bias=None)`, I guess it will still fall into this branch?

allo/ir/builder.py
Zhichenzzz: I think that this memref_copy is necessary maybe due to mapping memory operation, according to [this](https://discourse.llvm.org/t/rfc-linalg-add-concatenate-operation-to-linalg/2879/1) discussion and our test cases with assertion error results. Besides, it will generate copy ops when `mlir-opt` the mlir codes (in tensor dialect) with some passes.

allo/ir/builder.py
Zhichenzzz: You are exactly right. I will fix it.

allo/ir/builder.py
```diff
@@ -29,6 +29,7 @@
     TypeAttr,
     ArrayAttr,
     Attribute,
+    Type,
```
chhzh123: Probably it is better to rename this type to distinguish it from the AlloType

```python
from hcl_mlir.ir import Type as MLIRType
```

PR #89, opened 2023-09-29
[IR] Add passes to monitor memory usage

mmengjiadai:
## Description ##
This PR adds pass `monitor_memory_pass()` to passes.py and file test_monitor_memory.py to test the pass using simple cases.

### Problems ###
After lowering, arguments are stored in memory by `memref.alloc`. In order to minimize memory usage, we first need to determine the total usage of `memref.alloc`. 

### Proposed Solutions ###
We design pass to monitor the use of memory by traversing every operation in `intermediate_module`, which is the module generated right after going through lowering passes in `LLVMModule`.
First, we record zero constants in the list `zero_const`
```
# record zero constants
if isinstance(body_op, arith_d.ConstantOp):
                        dtype = body_op.type
                        if not isinstance(dtype, IndexType):
                            value = body_op.literal_value
                            if value == 0:
                                name = str(body_op).split("=", maxsplit=1)[0].strip()
                                zero_const.append(name)
```
Then, we record `memref.alloc` and storage by searching for `memref.store` in `scf.for`.
```
def find_storeop_in_forop(op):
        result = None
        for body_op in op.body.operations:
            if isinstance(body_op, memref_d.StoreOp):
                result = body_op
            elif isinstance(body_op, scf_d.ForOp):
                result_iter = find_storeop_in_forop(body_op)
                if result is None:
                    if result_iter is not None:
                        result = result_iter
                        break
                    raise NotImplementedError("No storeop found")
        return result
```
We format the results into a table using `tabulate` and return the string.

The argument `monitor_memory` is added to function `from_pytorch()` to monitor cases in allo/example.

### Examples ###
Running `allo/examples/mlp.py` with `monitor_memref=True` will generate file `mlp_mem_usage_1.txt`  with contents:
```
+-----------+----------+-------+-----------+------------+--------------+----------------------------------------------------------------------------+
|    name   |  shape   | dtype | mem(bits) | BRAM(18K)  | store counts |                                data storage                                |
+-----------+----------+-------+-----------+------------+--------------+----------------------------------------------------------------------------+
|   %alloc  | [30, 30] |  f32  |   28800   | 1638400.0  |      0       |                                                                            |
+-----------+----------+-------+-----------+------------+--------------+----------------------------------------------------------------------------+
|  %alloc_0 | [30, 30] |  f32  |   28800   | 1638400.0  |      1       |           %4 = memref.load %0[%arg1, %arg2] : memref<30x30xf32>            |
 +-----------+----------+-------+-----------+------------+--------------+----------------------------------------------------
...
+-----------+----------+-------+-----------+------------+--------------+----------------------------------------------------------------------------+
| %alloc_58 | [30, 30] |  f32  |   28800   | 1638400.0  |      0       |                                                                            |
+-----------+----------+-------+-----------+------------+--------------+----------------------------------------------------------------------------+
| Total(12) |          |       |   345600  | 19660800.0 |              | *data storage: data stored into an allocated memory. Doesn't include init. |
+-----------+----------+-------+-----------+------------+--------------+----------------------------------------------------------------------------+
```

## Checklist ##

- [x] Code is well-documented

### DIFF
--- a/allo/passes.py
+++ b/allo/passes.py
@@ -4,6 +4,7 @@
 
 import os
 import re
+import math
 import numpy as np
 from tabulate import tabulate
 
@@ -436,7 +437,7 @@ def monitor_memory_usage(intermediate_module):
                             mem_bits *= dim
                         data_bits = int(re.search(r"\d+", mem_dtype).group())
                         mem_bits *= data_bits
-                        bram = round(mem_bits / (18 * 1024), 2)
+                        bram = math.ceil(mem_bits / (18 * 1024))
                         store_count = 0
                         mem_alloc[alloc_name].append(
                             [mem_shape, mem_dtype, mem_bits, bram, store_count]
--- a/tests/test_monitor_memory.py
+++ b/tests/test_monitor_memory.py
@@ -21,7 +21,7 @@ def test_add_same_float():
     monitor_memory_table = monitor_memory_usage(mod.intermediate_module)
     print(monitor_memory_table)
     assert (
-        "| %alloc    | [32, 32] | f32     |       32768 |        1.78 | 2              |"
+        "| %alloc    | [32, 32] | f32     |       32768 |           2 | 2              |"
         in monitor_memory_table
     )
 
@@ -40,14 +40,14 @@ def test_add_const_float():
     monitor_memory_table = monitor_memory_usage(mod.intermediate_module)
     print(monitor_memory_table)
     assert (
-        "| %alloc_1  | [32, 32] | f32     |       32768 |        1.78 | 2              |"
+        "| %alloc_1  | [32, 32] | f32     |       32768 |           2 | 2              |"
         in monitor_memory_table
     )
 
 
 def test_calculate():
-    def kernel(A: float32[8], B: float32[8]) -> float32[8]:
-        c: float32[8] = 3
+    def kernel(A: float32[32], B: float32[32]) -> float32[32]:
+        c: float32[32] = 3
         return allo.exp(A * B + c)
 
     s = allo.customize(kernel, enable_tensor=True)
@@ -56,7 +56,7 @@ def test_calculate():
     monitor_memory_table = monitor_memory_usage(mod.intermediate_module)
     print(monitor_memory_table)
     assert (
-        "| %alloc_1 | [8]     | f32     |         256 |        0.01 | 2              |"
+        "| %alloc_1 | [32]    | f32     |        1024 |           1 | 2              |"
         in monitor_memory_table
     )
 
@@ -77,7 +77,7 @@ def test_gemm_grid_for():
     monitor_memory_table = monitor_memory_usage(mod.intermediate_module)
     print(monitor_memory_table)
     assert (
-        "| %alloc   | [32, 32] | i32     |       32768 |        1.78 | 1              |"
+        "| %alloc   | [32, 32] | i32     |       32768 |           2 | 1              |"
         in monitor_memory_table
     )
 


allo/passes.py
```diff
@@ -384,3 +394,217 @@ def update_generic_op(op, name, shape):
             raise NotImplementedError("Unsupported gelu shape")
     else:
         raise NotImplementedError("Unsupported function")
+
+
+# monitor tensor usage
+def get_tensor_usage(module):
```
chhzh123: Directly name the function `monitor_tensor_usage`

allo/passes.py
```diff
@@ -384,3 +394,217 @@ def update_generic_op(op, name, shape):
...
+    print("Getting tensor usage...")
```
chhzh123: Remove `print` in the pass

allo/passes.py
```diff
@@ -384,3 +394,217 @@ def update_generic_op(op, name, shape):
...
+                        dtype = body_op.type
+                        if isinstance(dtype, (IntegerType, F32Type)):
+                            value = body_op.literal_value
+                            if value == 0:
+                                name = str(body_op).split("=", maxsplit=1)[0].strip()
+                                zero_const.append(name)
+                    # record zero constants from casting
+                    elif isinstance(body_op, arith_d.SIToFPOp):
+                        name = str(body_op).split("=", maxsplit=1)[0].strip()
+                        zero_const.append(name)
```
chhzh123: Seems ad hoc here. What if the program has other `SIToFPOp`s?

allo/passes.py
```diff
@@ -384,3 +394,217 @@ def update_generic_op(op, name, shape):
...
+                    # record tensor.empty
+                    elif isinstance(body_op, tensor_d.EmptyOp):
+                        tensor_name = str(body_op).split("=", maxsplit=1)[0].strip()
```
chhzh123: I guess you can directly use `body_op.name` or sth like that? You can refer to MLIR's documentation. I don't think it is necessary to do string operation here

allo/passes.py
```diff
@@ -384,3 +394,217 @@ def update_generic_op(op, name, shape):
...
+    table.add_row(
+        [
+            "Total(" + str(total_tensor_count) + ")",
+            "",
+            "",
+            "",
+            "*other names: names\n of tensor after linalg",
+        ]
+    )
+    print(table)
```
chhzh123: return table and print outside

allo/passes.py
```diff
@@ -384,3 +394,217 @@ def update_generic_op(op, name, shape):
...
+
+
+# define a global variable to count the number of calls
+call_counter = {}
```
chhzh123: It will pollute this file if you define a global variable, which can be viewed by all the functions. I don't think it is needed here, since you are not using recursive functions.

allo/passes.py
```diff
@@ -384,3 +394,217 @@ def update_generic_op(op, name, shape):
...
+
+
+# monitor memory usage
+def get_mem_usage(module, enable_tensor=False):
```
chhzh123: Same. `monitor_mem_usage`

chhzh123: Also, are there any differences between this function and the `monitor_tensor_usage` function? I would suggest merging them into one

allo/frontend/pytorch.py
```diff
@@ -19,9 +19,17 @@
...
 
 
-def from_pytorch(model, example_inputs, verbose=False):
+def from_pytorch(
+    model,
+    example_inputs,
+    verbose=False,
+    enable_tensor=False,
+    monitor_tensor=False,
+    monitor_memref=False,
```
chhzh123: Merge into one argument

allo/customize.py
```diff
@@ -627,13 +627,14 @@ def compose(self, *schs):
                                     ("partition", [target] + list(args[1:]), kwargs)
                                 )
 
-    def build(self, target=None, mode=None, project=None):
+    def build(self, target=None, mode=None, project=None, monitor_memref=False):
```
chhzh123: No need to expose `monitor_memref` in the `build` function

allo/backend/llvm.py
```diff
@@ -52,7 +53,7 @@ def invoke_mlir_parser(mod: str):
 
 
 class LLVMModule:
-    def __init__(self, mod, top_func_name, ext_libs=None):
+    def __init__(self, mod, top_func_name, ext_libs=None, monitor_memref=False):
```
chhzh123: Same, no need to expose this interface. Users should directly call a pass

allo/passes.py
mmengjiadai: Perhaps I can verify if the value of SIToFPOp is 0？

allo/passes.py
mmengjiadai: I tried the attributes in the class definition and can't find one that works. Are there other ideas I can try?

allo/passes.py
mmengjiadai: I tested these functions on all cases in folder examples. Do I still need to add a file of testcases? I guess the two functions are similar but the details are different. Are you suggesting combining them and returning two tables? It took a long time when I ran these passes on cases like gp2.py. I'm not sure if merging them would cause unnecessary waiting time. For example, when I only want to monitor memrefs when enable_tensor=False.

allo/passes.py
chhzh123: Maybe. This one should be as specific as possible in order to eliminate false positive samples.

allo/passes.py
chhzh123: `body_op.operation.name`? You can check the class methods or the parent class methods

allo/passes.py
chhzh123: I mean you need to test the pass individually. For example, create a simple case (can be the one from the UW slide), and test whether the number of `store count` is expected.

chhzh123: In the unit test, no need to test a whole model like gpt2

chhzh123: So what are the differences between these two functions? The previous one counts the tensor store and the latter one counts the memref? I think when you dis/enable tensor, it has already determined whether the program contains memref/tensor

allo/passes.py
```diff
@@ -1,9 +1,15 @@
...
+# pylint: disable=no-name-in-module, unexpected-keyword-arg, no-value-for-parameter, too-many-nested-blocks
 
 import os
+import re
 import numpy as np
+
+try:
+    from prettytable import PrettyTable
+except ImportError:
+    pass
```
chhzh123: Use `tabulate` library instead

allo/passes.py
```diff
@@ -384,3 +394,216 @@ def update_generic_op(op, name, shape):
             raise NotImplementedError("Unsupported gelu shape")
     else:
         raise NotImplementedError("Unsupported function")
+
+
+# monitor memory usage
```
chhzh123: Delete comment

allo/passes.py
```diff
@@ -384,3 +394,216 @@ def update_generic_op(op, name, shape):
...
+                        dtype = body_op.type
+                        if isinstance(dtype, (IntegerType, F32Type)):
+                            value = body_op.literal_value
+                            if value == 0:
+                                name = str(body_op).split("=", maxsplit=1)[0].strip()
+                                zero_const.append(name)
+                    # record zero constants from casting
+                    elif isinstance(body_op, arith_d.SIToFPOp):
+                        match = re.search(r"arith\.sitofp\s(.*?):", str(body_op))
+                        extracted_str = match.group(1).strip()  # 去除首尾空格
```
chhzh123: Use English comments

allo/passes.py
```diff
@@ -384,3 +394,216 @@ def update_generic_op(op, name, shape):
...
+            "",
+            "",
+            "*other names: names\n of tensor after linalg",
+        ]
+    )
+    return table
+
+
+# monitor memory usage
+def monitor_memref_usage(module):
```
chhzh123: If a variable/function name has already conveyed the meaning, then there is no need to comment.

allo/passes.py
```diff
@@ -384,3 +394,216 @@ def update_generic_op(op, name, shape):
...
+    # find storeop in forop
+    def find_storeop_in_forop(op):
```
chhzh123: Same. The comment here does not provide new information.

tests/test_monitor_memory.py
```diff
@@ -0,0 +1,71 @@
...
+        C = B + 1
+        D = C + 1
+        E = D + 1
+        return E
+
+    s = allo.customize(kernel, enable_tensor=True)
+    print(s.module)
+    mod = s.build()
+    monitor_memory_table = monitor_memory_usage(s.module, mod.intermediate_module, True)
+    print(monitor_memory_table)
```
chhzh123: These test cases are almost the same as the first one, meaning they don't have good coverage of the code and may easily miss potential bugs. To enhance the robustness of the tests, I would suggest:
1. Add one example from https://dlsys.cs.washington.edu/pdf/lecture9.pdf
2. Test both `enable_tensor` and disable_tensor.
3. Instead of directly printing the memory table, use asserting to test if some of the memory counts are correct.
4. Use `pytest.mark.parametrize` to parameterize the `enable_tensor` flag and data type. See https://github.com/cornell-zhang/allo/blob/main/tests/test_linalg.py#L217

allo/passes.py
```diff
@@ -384,3 +394,216 @@ def update_generic_op(op, name, shape):
...
+        result = None
+        for body_op in op.body.operations:
+            if isinstance(body_op, memref_d.StoreOp):
+                result = body_op
+            elif isinstance(body_op, scf_d.ForOp):
```
chhzh123, line 401: You probably need to test whether the traditional `affine.for` loop can also be captured by your pass. For instance, you can try this basic example:
https://github.com/cornell-zhang/allo/blob/main/tests/test_schedule_compute.py#L10-L15

allo/passes.py
```diff
@@ -384,3 +394,216 @@ def update_generic_op(op, name, shape):
...
+                result_iter = find_storeop_in_forop(body_op)
+                if result is None:
+                    if result_iter is not None:
+                        result = result_iter
```
chhzh123, line 405: No `break` after this statement?

tests/test_monitor_memory.py
```diff
@@ -0,0 +1,97 @@
...
+    mod = s.build()
+    monitor_memory_table = monitor_memory_usage(mod.intermediate_module)
+    pattern = (
+        r"\|\s+%alloc\s+\|\s+\S+\s+\S+\s+\|\s+\S+\s+\|\s+\S+\s+\|\s+\S+\s+\|\s+2\s+\|"
+    )
+    match = re.search(pattern, monitor_memory_table)
+    assert (
+        match.group(0).strip()
+        == "| %alloc    | [32, 32] | f32     |       32768 | 1.86414e+06 | 2              |"
+    )
```
chhzh123: I don't quite get it. Why not directly assert `| %alloc ...` in `monitor_memory_table`?

chhzh123: If using pattern matching, then it supposes to return only `%alloc` and `2` instead of a whole string

allo/passes.py
```diff
@@ -384,3 +389,112 @@ def update_generic_op(op, name, shape):
...
+                        mem_alloc[alloc_name] = []
+                        mem_type = body_op.result.type
+                        mem_shape = mem_type.shape
+                        mem_dtype = str(mem_type.element_type)
+                        mem_bits = 1
+                        for dim in mem_shape:
+                            mem_bits *= dim
+                        data_bits = int(re.search(r"\d+", mem_dtype).group())
+                        mem_bits *= data_bits
+                        bram = round(mem_bits / 18 * 1024, 2)
```
chhzh123: Your BRAM calculation is not correct. Should be `mem_bits/(18*1024)`.

allo/passes.py
```diff
@@ -384,3 +389,112 @@ def update_generic_op(op, name, shape):
...
+                        mem_alloc[alloc_name] = []
+                        mem_type = body_op.result.type
+                        mem_shape = mem_type.shape
+                        mem_dtype = str(mem_type.element_type)
+                        mem_bits = 1
+                        for dim in mem_shape:
+                            mem_bits *= dim
+                        data_bits = int(re.search(r"\d+", mem_dtype).group())
+                        mem_bits *= data_bits
+                        bram = round(mem_bits / (18 * 1024), 2)
```
chhzh123: Check out this [slide](https://www.csl.cornell.edu/courses/ece6775/pdf/lecture04.pdf). BRAM is a hardware unit, so it can only be an integer.

PR #120, opened 2023-12-04
[Bugfix] Support schedule composition for `reuse_at`

zzzDavid:
## Description ##

Schedule composition did not support `reuse_at`, this PR adds this support for kernels such as `jacobi-2d`.

### Problems ###
An example:
```python
def test_reuse_at_compose():
    def reuse_blur_x_y(A: int32[10, 10]) -> int32[8, 8]:
        B: int32[8, 8] = 0
        for y, x in allo.grid(8, 8):
            B[y, x] = A[y, x] + A[y + 1, x + 1] + A[y + 2, x + 2]
        return B

    s = allo.customize(reuse_blur_x_y)
    RB_y = s.reuse_at(s.A, "y")
    RB_x = s.reuse_at(RB_y, "x")

    def top_kernel(A: int32[10, 10]) -> int32[8, 8]:
        B = reuse_blur_x_y(A)
        return B * 2

    s_top = allo.customize(top_kernel)
    s_top.compose(s)
```

This will report the following error:
```
    sch.compose(s)    
  File "/work/shared/users/phd/nz264/allo/allo/customize.py", line 110, in wrapper
    res = fn(*args, **kwargs)
  File "/work/shared/users/phd/nz264/allo/allo/customize.py", line 752, in compose
    primitive_func.__wrapped__(self, *args, **kwargs)
  File "/work/shared/users/phd/nz264/allo/allo/customize.py", line 523, in reuse_at
    raise RuntimeError("Reuse buffer not found")
RuntimeError: Reuse buffer not found
```
Because the compiler is looking for target reuse buffer only at the top level function.

### Proposed Solutions ###

Look for target reuse buffer in all functions, not just the top-level function. This is implemented in `find_reuse_buffers`. 

### Examples ###
The test case is added in `test_schedule_compose.py::test_reuse_at_compose`. `jacobi-2d` will also be added in the PolyBench PR.

## Checklist ##

- [x] Code is well-documented

### DIFF
--- a/tests/test_schedule_compose.py
+++ b/tests/test_schedule_compose.py
@@ -459,5 +459,4 @@ def test_reuse_function_2():
 
 
 if __name__ == "__main__":
-    # pytest.main([__file__])
-    test_two_reuse_at_compose()
+    pytest.main([__file__])


tests/test_schedule_compose.py
```diff
@@ -378,4 +459,5 @@ def top(inp: int32[M, N]) -> int32[M, N]:
 
 
 if __name__ == "__main__":
-    pytest.main([__file__])
+    # pytest.main([__file__])
+    test_two_reuse_at_compose()
```
chhzh123: Recover the first line

zzzDavid: Thanks, yes that was for my local testing, not meant to be pushed. I've reverted this change

zzzDavid: Please merge once the tests complete

PR #164, opened 2024-07-07
Corrected vivado_hls g++ version check

bibo-msft:
## Description ##
Corrected the parsing of the version returned from g++ version.

### Problems ###
G++ 11.4.0 was judged as a "too old" version.

### Proposed Solutions ###
11 * 10 + 4 >= 48 will be working with this change.
4 * 10 + 8 >= 48 will be working with this change.

### Examples ###
(Please provide an example of the input program and the expected behavior, e.g., the generated IR.)

## Checklist ##

- [ ] Code is well-documented

### DIFF
--- a/allo/backend/hls.py
+++ b/allo/backend/hls.py
@@ -316,7 +316,7 @@ class HLSModule:
             ), f"cannot find {self.platform} on system path"
             ver = run_process("g++ --version", r"\d+\.\d+\.\d+")[0].split(".")
             assert (
-                int(ver[0]) * 4 + int(ver[1]) >= 48
+                int(ver[0]) * 10 + int(ver[1]) >= 48
             ), f"g++ version too old {ver[0]}.{ver[1]}.{ver[2]}"
 
             cmd = f"cd {self.project}; make "
--- a/allo/harness/vivado/run.tcl
+++ b/allo/harness/vivado/run.tcl
@@ -21,7 +21,7 @@ add_files -tb host.cpp -cflags "-std=gnu++0x"
 
 open_solution "solution1"
 # Use Zynq device
-set_part {xczu3eg-sbva484-1-i}
+set_part {xc7z020clg484-1}
 
 # Target clock period is 10ns
 create_clock -period 10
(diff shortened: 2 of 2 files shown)


allo/backend/hls.py
```diff
@@ -314,9 +314,9 @@ def __call__(self, *args, shell=True):
             assert (
                 os.system(f"which {self.platform} >> /dev/null") == 0
             ), f"cannot find {self.platform} on system path"
-            ver = run_process("g++ --version", r"\d\.\d\.\d")[0].split(".")
+            ver = run_process("g++ --version", r"\d+\.\d+\.\d+")[0].split(".")
             assert (
-                int(ver[0]) * 10 + int(ver[1]) >= 48
+                int(ver[0]) * 4 + int(ver[1]) >= 48
```
chhzh123: Maybe we can leave it as it is. `11*10 + 4>=48` also works for your case.

allo/harness/vivado/run.tcl
```diff
@@ -21,7 +21,7 @@ add_files -tb host.cpp -cflags "-std=gnu++0x"
 
 open_solution "solution1"
 # Use Zynq device
-set_part {xc7z020clg484-1}
+set_part {xczu3eg-sbva484-1-i}
```
chhzh123: We will be providing an interface for passing in the target FPGA board name, so please keep this tcl file unchanged in this PR.

chhzh123 (commented):
Thanks for contributing to Allo and fixing the issue! Could you please clarify the following comments so that we can merge your PR?

allo/backend/hls.py
bibo-msft: What is the original minimum required g++ version?

My default Ubuntu 22.04 has version g++ 11.4.0. And 11 * 4 + 4 = 48.
And for g++ 12.x.y, 12 * 4 + x >= 48 also works. 
That's why I guessed the original factor should be 4, rather than 10.

chhzh123: This file originated from an old project, so the minimal requirement is just 4.8. If you change the factor to 4, then `4 * 4 + 8 >= 48` does not work :)

chhzh123:
Thanks for revising the PR!

Issue #176, opened 2024-09-09
[BUG] NaN Wrong Results of Scaled Dot Product Attention due to Softmax Overflow


silvenachen:
**Describe the bug**
When I'm calling scaled_dot_product_attention() function from Allo pre-build library in a larger kernel, the intermediate results are incorrect and yields NaNs. The issue stems from numerical overflow during the softmax computation, where large values in the exponentiation step cause overflow, leading to NaNs.

**To Reproduce**
```
import allo
from allo.ir.types import float32
import numpy as np

def test_allo_sdp():
    from allo.library.nn import scaled_dot_product_attention
    L, D = 1024, 512
    Q = np.random.randn(L, D).astype(np.float32)
    K = np.random.randn(L, D).astype(np.float32)
    V = np.random.randn(L, D).astype(np.float32)
    Q = 1000*Q
    K = 1000*K
    V = 1000*V
    H = 8
    Q_np = Q.reshape((L, H, D//H))
    K_np = K.reshape((L, H, D//H))
    V_np = V.reshape((L, H, D//H))

    s = allo.customize(scaled_dot_product_attention, instantiate=[float32, H, L, D, 8, 8])
    mod = s.build()
    allo_out = mod(Q, K, V)
    print(allo_out)

if __name__ == "__main__":
    test_allo_sdp()
```

**Buggy output**
[[nan nan nan ... nan nan nan]
 [nan nan nan ... nan nan nan]
 [nan nan nan ... nan nan nan]
 ...
 [nan nan nan ... nan nan nan]
 [nan nan nan ... nan nan nan]
 [nan nan nan ... nan nan nan]]

**Expected behavior**
The scaled_dot_product_attention should function and gives correct numerical output even when input is large.

**Additional context**
The cause of NaNs is due to the overflow of softmax operation, as the original Allo function can not handle larger input numbers and results in overflow.

chhzh123:
@silvenachen Thanks for raising this issue! Have you fixed the softmax function? If so, could you help create a PR for Allo?

silvenachen:
> @silvenachen Thanks for raising this issue! Have you fixed the softmax function? If so, could you help create a PR for Allo?

Yes, I think I have fixed this issue for both softmax and attention computation. How could I help to maybe update the library?

PR #180, opened 2024-09-10
[Bugfix] Resolve softmax and sdp overflow for large values

silvenachen:
## Description ##
The PR solves the issue related to numerical overflow during softmax calculation and the scaled_dot_product_attention that utilizes it. In addition, a similar overflow error occurs in the numpy implementation used in `allo/tests/test_nn.py` that generate golden results for testing. The PR fixes this issue by adding a correct `np_softmax()` function in the testing script.

### Problems ###
- **Softmax Overflow**: The softmax calculation could result in overflow when handling large input values, causing NaNs in hte output.
- **Incorrect Numpy Softmax Implementation**: In the original test script, `np_out = np.exp(inp) / np.exp(inp).sum(axis=1, keepdims=True)` will cause similar overflow problems in the exponentiation computation.

### Proposed Solutions ###
The softmax function has been updated to subtract the maximum value in each row before exponentiation to cope with larger values correctly.

### Examples ###
```
(allo) sl3558@brg-zhang-xcel:/work/shared/users/ugrad/sl3558/mdlm/dit_numpy_allo$ pytest allo/tests/test_nn.py::test_softmax
========================================================= test session starts =========================================================
platform linux -- Python 3.12.4, pytest-8.2.2, pluggy-1.5.0
rootdir: /work/shared/users/ugrad/sl3558/mdlm/dit_numpy_allo/allo
collected 1 item                                                                                                                      

allo/tests/test_nn.py .                                                                                                         [100%]

========================================================== 1 passed in 2.91s ==========================================================
```

## Checklist ##

- [ ] Code is well-documented

### DIFF
--- a/allo/library/nn.py
+++ b/allo/library/nn.py
@@ -13,17 +13,17 @@ def softmax[Ty, L](X: "Ty[L, L]") -> "Ty[L, L]":
     S: Ty[L] = 0.0
 
     for i, j in dsl.grid(L, L, name="row_max"):
-        if X[i,j] > M[i]:
+        if X[i, j] > M[i]:
             M[i] = X[i, j]
 
-    #compute exp and sum
+    # compute exp and sum
     for i, j in dsl.grid(L, L, name="exp_sum"):
         E[i, j] = dsl.exp(X[i, j] - M[i])
         S[i] += E[i, j]
-    
+
     for i, j in dsl.grid(L, L, name="update"):
         Z[i, j] = E[i, j] / S[i]
-    
+
     return Z
 
 
--- a/allo/library/nn.py
+++ b/allo/library/nn.py
@@ -6,17 +6,24 @@ from .. import dsl
 from .systolic import systolic
 
 
-def softmax[Ty, D](X: "Ty[D, D]") -> "Ty[D, D]":
-    Z: Ty[D, D]
-    E: Ty[D, D]
-    S: Ty[D] = 0.0
+def softmax[Ty, L](X: "Ty[L, L]") -> "Ty[L, L]":
+    Z: Ty[L, L]
+    E: Ty[L, L]
+    M: Ty[L] = -10000.0
+    S: Ty[L] = 0.0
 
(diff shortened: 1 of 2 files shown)


allo/library/nn.py
```diff
@@ -6,17 +6,24 @@
...
-    Z: Ty[D, D]
-    E: Ty[D, D]
-    S: Ty[D] = 0.0
-
-    for i, j in dsl.grid(D, D, name="exp_sum"):
-        E[i, j] = dsl.exp(X[i, j])
+def softmax[Ty, L](X: "Ty[L, L]") -> "Ty[L, L]":
+    Z: Ty[L, L]
+    E: Ty[L, L]
+    M: Ty[L] = -10000.0
```
chhzh123: Maybe use a larger constant number, say `-0x3f3f3f3f.0` (a large enough number smaller within the range of int32)

tests/test_nn.py
```diff
@@ -51,14 +51,20 @@ def test_linear_float():
...
+    e_x = np.exp(x - x_max)
+    return e_x / np.sum(e_x, axis=axis, keepdims=True)
+
+
 def test_softmax():
     from allo.library.nn import softmax
 
     s = allo.customize(softmax, instantiate=[float32, 8])
     mod = s.build()
     inp = np.random.randn(8, 8).astype(np.float32)
```
chhzh123, line 65: Can you also use a larger test case? I remember you used some large random tensors in your own test.

chhzh123 (commented):
Thanks for contributing to Allo! Just some minor issues

allo/library/nn.py
silvenachen: I could not directly use -0x3f3f3f3f.0, but I used another value which I believe is also negative enough,

tests/test_nn.py
silvenachen: Sure! I added `inp = 1000 * inp` to create larger tensor test cases.

chhzh123 (approved):
LGTM. Thanks! This PR fixes #176.

PR #181, opened 2024-09-11
[Bugfix] Add scf ForOp support and nested loop search in dataflow and pipeline

AdrianLiu00:
## Description ##
1. Add scf ForOp support for dataflow() and pipeline() in customize.py;
2. Add nested loop search support for dataflow() in customize.py

### Problems ###
1. Absence of scf ForOp support in some functions.
2. Fail to search the inner loop when it's deeply nested when executing dataflow().

### Proposed Solutions ###
Modify the recursive searching in nested loop.

### Examples ###
Provided in the test_schedule_stream.py.

## Checklist ##

- [√] PR's title starts with a category (e.g. [Bugfix], [IR], [Builder], etc)
- [√] Changes are complete (i.e. I finished coding on this PR)
- [√] All changes have test coverage (It would be better to provide ~2 different test cases to test the robustness of your code)
- [ ] Code is well-documented

### DIFF
--- a/allo/customize.py
+++ b/allo/customize.py
@@ -433,11 +433,11 @@ class Schedule:
         band_name = band_name.split(":")[1]
         cnt = 0
 
-        def locateLoop(op):
+        def locate_loop(op):
             nonlocal cnt
             for ope in op.body.operations:
                 if isinstance(ope, (scf_d.ForOp, affine_d.AffineForOp)):
-                    locateLoop(ope)
+                    locate_loop(ope)
             if (
                 "loop_name" in op.attributes
                 and op.attributes["loop_name"].value == loop_name
@@ -451,7 +451,7 @@ class Schedule:
                     "op_name" in op.attributes
                     and op.attributes["op_name"].value == band_name
                 ):
-                    locateLoop(op)
+                    locate_loop(op)
 
         if cnt == 0:
             raise RuntimeError(f"Dataflow loop {band_name}.{loop_name} not found")
--- a/allo/customize.py
+++ b/allo/customize.py
@@ -32,6 +32,7 @@ from hcl_mlir.dialects import (
     hcl as hcl_d,
     memref as memref_d,
     affine as affine_d,
+    scf as scf_d,
     arith as arith_d,
     func as func_d,
 )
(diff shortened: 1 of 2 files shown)


allo/customize.py
```diff
@@ -435,24 +433,25 @@ def dataflow(self, axis):
         band_name = band_name.split(":")[1]
         cnt = 0
 
-        # TODO: Fix deep nested
-        def DFS(op):
+        def locateLoop(op):
```
chhzh123: Follow [Python naming convention](https://peps.python.org/pep-0008/): `locate_loop`

tests/test_schedule_stream.py
```diff
@@ -120,5 +120,46 @@ def top(A: T[10, 20]) -> T[10, 20]:
...
+        for i1 in range(times_out, name="out1"):
+            for i2 in range(A, name="out2"):
+                for j in range(B, name="target"):
+                    for k in range(N, name="in1"):
+                        C[k] = C[k] + 1
+                    for k in range(N, name="in2"):
+                        C[k] = C[k] - 1
+
+    s1 = allo.customize(case1, instantiate=[8])
+    target_loop = s1.get_loops(s1.top_func_name)["out1"]["j"]
```
chhzh123, line 158: Why "j" here would work? I'm still debating whether the "j" loop should be exposed at the outer loop level or should access it by `["out1"]["i2"]["j"]`

chhzh123 (commented):
Thanks for the PR! I see you didn't use the `band_name`. I'm wondering if it will be a problem if there are multiple for loops using the same variable name. We need some mechanism to distinguish between loops.

AdrianLiu00:
Yeah, it's all about the way that customize() get access to the inner loop. Since the nested loops are now all exposed to the outermost loop, so I just use `band_name` in the condition to start the recursive locateLoop() instead inside the traversal procedure.
```
        for op in func.entry_block.operations:
            if isinstance(op, (scf_d.ForOp, affine_d.AffineForOp)):
                if (
                    "op_name" in op.attributes
                    and op.attributes["op_name"].value == band_name
                ):
                    locateLoop(op)
```

In my opinion, it's more intuitive for user to access inner loop by `["out1"]["i2"]["j"]`. And it will make locate the specific loop esaier than the traversal procedure now. But this modification could be possibly thorough, because all previously relevant part should be modified.

We can discuss further about how to improve it. Also, the problem to identify the loops using the same variable name is supposed to address as well maybe.

chhzh123:
Ah sorry, I didn't notice you used `band_name`. You can fix the naming issue and go ahead. For the loop access, we can open a new issue to discuss.

PR #198, opened 2024-09-22
Adding essential header libraries for compiling DPC++ file using DPC+…

JackXinhangGUO:
## Description ##
Updating front-end code to support Intel HLS

### Problems ###
The Intel HLS platform has not been an option for users to choose as a platform

### Proposed Solutions ###
Updated hls.py and customize.py to support Intel HLS and added the option mode to either output HLS code with or without compiling the HLS code using the DPC++ compiler if 'mode' is None, the output file will not be generated and Intel HLS require 'project' argument must not be None, and 'mode' must be specified when 'project' is specified.

### Future Work ###
the DPC++ compiler on the server must be updated to 2024.x.xx version in order to use the Intel HLS function.

### DIFF
--- a/allo/backend/hls.py
+++ b/allo/backend/hls.py
@@ -222,7 +222,13 @@ class HLSModule:
             copy_build_files(self.top_func_name, project, mode, platform=platform)
             copy_ext_libs(ext_libs, project)
             if self.platform == "vitis_hls":
-                assert self.mode in {"csim", "sw_emu", "hw_emu", "hw"}, "Invalid mode"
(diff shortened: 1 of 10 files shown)


chhzh123 (commented):
Several issues after an initial look:
1. Did you copy these files from some Intel libraries? Please provide references and do not remove the license file.
2. Please only keep necessary header files that are used for compilation and remove others.
3. The PR description also needs to be updated.
4. Need to update the submodule commit as well to reflect your latest changes in the hcl-dialect folder.

PR #206, opened 2024-09-29
Updating front-end code to accommodate Intel HLS and add library for …

JackXinhangGUO:
…compiling HLS code to bitstream for Intel FPGA

## Description ##
Updated front-end code to accommodate Intel HLS and add library for compiling HLS code to bitstream for Intel FPGA

### Problems ###
The Intel HLS option was not available for the user

### Proposed Solutions ###
updated the code to allow users to choose a mode to output Intel HLS code in the command line, output source HLS as a file, or compile the HLS code to an Intel FPGA emulator .exe file. added library under harness/intel_oneapi to support DPC++ compilation for FPGA.


allo/backend/hls.py
```diff
@@ -168,6 +170,39 @@ def __init__(
...
+
+            if platform == "ihls":
+                assert mode in {
+                    "fpga_emulator",
+                    "source_file_only",
+                    None,
+                }, "Invalid mode"
+
+                if project is not None:
+                    filename = f"{project}"
```
chhzh123: filename is different from `project`. For example, users can specify the `project` as a nested folder `test/out.prj`.

allo/backend/hls.py
```diff
@@ -168,6 +170,39 @@ def __init__(
...
+                else:
+                    raise RuntimeError(
+                        "Error: if platfrom is ihls, 'project' argument must not be None."
+                    )
+
+                if mode == "fpga_emulator":
+                    result = subprocess.run(
+                        [
+                            f" icpx -fintelfpga -DFPGA_EMULATOR .//{filename} -o {filename}.exe"
```
chhzh123: Use linux instruction. Also, how do you handle the non-emulator situations?

allo/backend/hls.py
```diff
@@ -168,6 +170,39 @@ def __init__(
...
+                        check=False,
+                    )
+                    print(result.stdout)
+                elif mode == "source_file_only":
+                    print(
+                        f"Generated Intel HLS source file kernel.cpp has been created successfully in your current directory under '{filename}' folder."
+                    )
+                    print(
+                        "mode has been set to source_file_only, the output will only be the souce intel HLS code"
+                    )
```
chhzh123: I don't think we need this mode. Users can always access the source code from the built folder

allo/backend/hls.py
```diff
@@ -278,8 +318,12 @@ def __init__(
...
             with open(f"{project}/kernel.cpp", "w", encoding="utf-8") as outfile:
                 outfile.write(self.hls_code)
-            with open(f"{project}/host.cpp", "w", encoding="utf-8") as outfile:
-                outfile.write(self.host_code)
+            if platform.lower() != "ihls":
+                with open(f"{project}/host.cpp", "w", encoding="utf-8") as outfile:
+                    outfile.write(self.host_code)
+            else:
+                print("Platform is intel HLS; skipping generation of host.cpp.")
+
```
chhzh123: The host code is also set as empty. No need to skip the generation. Otherwise, you'd better change the name of `kernel.cpp` as this one has already contained both host and device code.

allo/customize.py
```diff
@@ -663,11 +663,17 @@ def build(self, target=None, mode=None, project=None, configs=None):
...
                 ext_libs=self.ext_libs,
             )
-        if target in {"vhls", "vivado_hls", "vitis_hls"}:
+        if target in {"vhls", "vivado_hls", "vitis_hls", "ihls"}:
+            if target == "ihls":
+                platform = "ihls"
+            elif target == "vitis_hls":
+                platform = "vitis_hls"
+            else:
+                platform = "vivado_hls"
```
chhzh123: "vivado_hls" if target == "vhls" else target

chhzh123 (commented):
Thanks for contributing. It seems you're not using the included headers for now. I would suggest removing the `intel_oneapi` folder. Also, some test cases using `ihls` should be included. Please refer to this [test file](https://github.com/cornell-zhang/allo/blob/main/tests/test_vhls.py) for details.

chhzh123:
As the LLVM version has changed, this PR is no longer available. #277 implements the latest Intel HLS support.

PR #236, opened 2024-10-28
Remove unnecessary stream arguments and tag stream type in dataflow

EthanMeng324:
In dataflow implementation, unused stream arguments for each kernel, and the corresponding declaration will be examined and removed. For Tapa codegen, a tagging strategy for the type of stream (input, output, general) is also implemented in dataflow.


chhzh123 (commented):
Mostly LGTM. Can you create one or two simple test cases under `tests/dataflow/test_unit.py` to make sure there are no redundant streaming arguments generated? You can just do some simple pattern matching on the generated MLIR/C++ code.

EthanMeng324:
> Mostly LGTM. Can you create one or two simple test cases under `tests/dataflow/test_unit.py` to make sure there are no redundant streaming arguments generated? You can just do some simple pattern matching on the generated MLIR/C++ code.

I'm kind of confused how to write the pattern matching. I mean if we are using the df.kernel decorator, we can only access the function in the user level, and can't see the exact kernel implementation underneath.

chhzh123:
After #235, users can explicitly call the `df.build` function and pass in the target, just like [this example](https://github.com/cornell-zhang/allo/blob/main/tests/dataflow/test_tiled_systolic.py#L64). You can use the `vhls` backend to generate the C++ code and then test whether the number of arguments is expected, similar to [this kind of testing](https://github.com/cornell-zhang/allo/blob/main/tests/test_vhls.py#L130).

EthanMeng324:
> After #235, users can explicitly call the `df.build` function and pass in the target, just like [this example](https://github.com/cornell-zhang/allo/blob/main/tests/dataflow/test_tiled_systolic.py#L64). You can use the `vhls` backend to generate the C++ code and then test whether the number of arguments is expected, similar to [this kind of testing](https://github.com/cornell-zhang/allo/blob/main/tests/test_vhls.py#L130).

I created a simple test to check the number of arguments, it might be a little bit hard in python level to check if the exact arguments is placed correctly in each kernel.

tests/dataflow/test_unit.py
```diff
@@ -0,0 +1,84 @@
...
+    if hls.is_available("vitis_hls"):
+        gemm(A, B, C)
+        np.testing.assert_allclose(C, np.dot(A, B), atol=1e-5)
+        mod = df.build(gemm, target="vitis_hls", mode="csim", project="top.prj")
+        code = mod.hls_code
+        check_function_arguments(code, "gemm_0_0", 3)
+        check_function_arguments(code, "gemm_0_1", 4)
+        check_function_arguments(code, "gemm_1_0", 4)
+        check_function_arguments(code, "gemm_1_1", 7)
+        print("Passed!")
```
chhzh123: Since the CI server does not install Vitis HLS, can you just call `df.build(gemm, target="vhls")` and check the function arguments?

tests/dataflow/test_unit.py
```diff
@@ -0,0 +1,84 @@
...
+import allo.backend.hls as hls
+import re
+import numpy as np
+
+# M, N, K = 512, 512, 512
+# Mt, Nt = 16, 16
+M, N, K = 16, 16, 16
+Mt, Nt = 1, 1
+# M, N, K = 4, 4, 4
+# Mt, Nt = 1, 1
```
chhzh123: Maybe just use the `4,4,4` setting and remove others.

chhzh123:
The test looks great. Just changing the target should be perfect.

EthanMeng324:
> The test looks great. Just changing the target should be perfect.

All fixed!

PR #240, opened 2024-10-31
Remove unused kernel in dataflow

EthanMeng324:
Follow up from https://github.com/cornell-zhang/allo/pull/236 , The tiled systolic implementation will generate unused kernels using dataflow, like "gemm_0_0". This may work fine in vitis, but tapa.invoke will generate some clock error when dealing with empty kernel. Thus, I used "canonicalize" pass to remove the useless op and remove all empty kernels and corresponding callOp. The test_unit is also updated to check this behavior.


tests/dataflow/test_unit.py
```diff
@@ -4,32 +4,38 @@
...
+    # A[Mt, K] * B[K, Nt] = C[Mt, Nt]
     i, j = df.get_pid()
     in_A: Stream[int32] = df.pipe(src=(i, j - 1), dst=(i, j))
     in_B: Stream[int32] = df.pipe(src=(i - 1, j), dst=(i, j))
     out_A: Stream[int32] = df.pipe(src=(i, j), dst=(i, j + 1))
     out_B: Stream[int32] = df.pipe(src=(i, j), dst=(i + 1, j))
     for m in range(M // Mt):
         for n in range(N // Nt):
-            with allo.meta_if(i == 0 and j == 0):
+            with allo.meta_if(i in [0, Mt + 1, Nt + 1] and j in [0, Mt + 1, Nt + 1]):
```
chhzh123: Use set instead. For example, `{0, Mt + 1, Nt + 1}`

allo/dataflow.py
```diff
@@ -114,21 +115,46 @@ def tag_stream_type(func, func_name, stream_info, start_index):
...
+        ):
+            func_op.erase()
+            del func_names[i]
+
+
+def _build_top(s_top, input_types, stream_info, func_names):
     """
     s_top: schedule of top-level function
     input_types: top-level function input types
     stream_info: {(src, dst): stream_types} (TODO: support more than one stream)
```
chhzh123, line 135: Add `func_name` meaning

chhzh123 (commented):
Mostly LGTM. One question: What kind of operations do the `canonicalize` pass remove? The operations without usage, just like the empty `ForOp`?

EthanMeng324:
> canonicalize

Actually a lot of stuff, like constant folding, common subexpression elimination. Basically many optimization passes of a general compiler in LLVM IR phase. I see we also use that for llvm module.

EthanMeng324:
> Mostly LGTM. One question: What kind of operations do the `canonicalize` pass remove? The operations without usage, just like the empty `ForOp`?

Actually a lot of stuff, like constant folding, common subexpression elimination. Basically many optimization passes of a general compiler in LLVM IR phase. I see we also use that for llvm module.

chhzh123 (approved):
LGTM. Thx!

PR #242, opened 2024-11-01
[Builder] Add `index` from/to `Float`/`Fixed`/`UFixed` type casting

### LINKED ISSUE #241: [BUG] "Casting between unsupported types" error for fixed type and index type
**Describe the bug**
I'm trying to write something that uses a Fixed(32, 24) type and a Fixed(64, 48) type for data. In a loop, I have a variable step that increments in the loop and is of type Index. I'm trying to perform an operation (add) with the Fixed(32, 24) type and the Index type, and store the result in the Fixed(64, 48) variable. However, I get the error "[Data Type] Casting between unsupported types. src type: index, dst type: fixed(58, 24)". I tried changing the type of the destination variable to Fixed(58, 24), but that resulted in the same error, so I'm guessing that the issue is in the add operation between the Fixed(32, 24) type and the Index type. However, this doesn't seem to strictly be an add issue either, since originally it was a shift operation, but that still gave the same error.

**To Reproduce**
```
import allo                                                                                                             
from allo.ir.types import int32, bool, index, Fixed, float32                                                                               

theta_type = Fixed(32, 24)                                             
mult_type = Fixed(64, 48)                                                                    

def test(theta: theta_type):                                                                                                
    current_sin: theta_type = 0.0                                                                     
    for step in range(20):         
        mult_intermed_cos: mult_type = 0.0                                 
        mult_intermed_cos = current_sin + step 

s = allo.customize(test)       
print(s.module)                                          
mod = s.build(target="llvm")     
```

**Buggy output**
raise DTypeError(allo._mlir.exceptions.DTypeError: [Data Type] Casting between unsupported types. src type: index, dst type: fixed(58, 24)

**Expected behavior**
Expect the program to run correctly, and for the types to automatically cast so that the correct result is stored in mult_intermed_cos with the Fixed(64,48) datatype.

zzzDavid:
## Description ##
This PR introduces expanding support for casting between `Index`, `Fixed`, and `UFixed` types in the Allo IR. This enhancement allows for seamless conversions between these types, which broadens the flexibility of type handling in the IR layer.

### Problems ###
Previously, the Allo IR lacked direct support for conversions between `Index` and `Fixed/UFixed` types, limiting the flexibility in scenarios where such conversions are needed for operations. Attempting these conversions required workarounds or was unsupported, creating friction in code requiring dynamic type handling.

### Proposed Solutions ###
- Added `Index <-> Fixed/UFixed` single-step conversions in the `cast_map` within `ASTTransformer`.
- Implemented new casting operations `IntToFixedOp` and `FixedToIntOp` as intermediary conversions to facilitate direct type transformations.
- Updated the `IndexCastOp` to handle `Index -> Fixed/UFixed` conversions.

### Examples ###
Here’s a snippet of the updated behavior:

1. **Index to Fixed/UFixed Casting**:
   ```python
   def kernel(a: index) -> Fixed:
       a_fixed: Fixed = a
       return a_fixed
   ```

2. **Fixed/UFixed to Index Casting**:
   ```python
   def kernel(a: Fixed) -> index:
       a_idx: index = a
       return a_idx
   ```

Both conversions now produce correct IR output without errors, validating the new casting support.

## Checklist ##

- [x] PR's title starts with a category (e.g., [IR], [Builder])
- [x] Changes are complete (i.e., finished coding on this PR)
- [x] All changes have test coverage, with cases for both `Index -> Fixed/UFixed` and `Fixed/UFixed -> Index` conversions
- [x] Code is well-documented

### DIFF
--- a/allo/ir/builder.py
+++ b/allo/ir/builder.py
@@ -356,6 +356,7 @@ class ASTTransformer(ASTBuilder):
         if type(res_type) is type(src_type) and res_type == src_type:
             return op
 
+        # Single-step type conversions
         cast_map = {
             # Index <-> UInt/Int
             (Int, Index): arith_d.IndexCastOp,
@@ -367,9 +368,6 @@ class ASTTransformer(ASTBuilder):
             (UInt, Float): arith_d.UIToFPOp,
             (Float, Int): arith_d.FPToSIOp,
             (Float, UInt): arith_d.FPToUIOp,
-            # FP to Index is not supported in MLIR
-            # (Float, Index): RuntimeError,
-            # (Index, Float): RuntimeError,
             # Float <-> Fixed/UFixed
             (Float, Fixed): allo_d.FloatToFixedOp,
             (Float, UFixed): allo_d.FloatToFixedOp,
@@ -402,6 +400,16 @@ class ASTTransformer(ASTBuilder):
                 IntegerType.get_signless(32), op.result, ip=ctx.get_ip()
             )
             opcls = arith_d.SIToFPOp  # proceed to build cast to float
+        elif isinstance(src_type, Index) and isinstance(res_type, (Fixed, UFixed)):
+            op = arith_d.IndexCastOp(
+                IntegerType.get_signless(32), op.result, ip=ctx.get_ip()
+            )
+            opcls = allo_d.IntToFixedOp  # proceed to build cast to float
+        elif isinstance(src_type, (Fixed, UFixed)) and isinstance(res_type, Index):
+            op = allo_d.FixedToIntOp(
+                IntegerType.get_signless(32), op.result, ip=ctx.get_ip()
+            )
+            opcls = arith_d.IndexCastOp
         elif isinstance(src_type, (Int, UInt)) and isinstance(res_type, (Int, UInt)):
             if src_type.bits > res_type.bits:
                 opcls = arith_d.TruncIOp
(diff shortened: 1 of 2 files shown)


tests/test_types.py
```diff
@@ -42,6 +43,44 @@ def kernel(a: int32) -> float32:
     assert mod(1) == kernel(1)
 
 
+def test_index_fixed_casting():
+    def test_one_cast(fixed):
+        def kernel(a: fixed) -> float32:
```
chhzh123: where `a` is used here?

allo/ir/builder.py
```diff
@@ -367,9 +368,6 @@ def build_cast_op(ctx, op, src_type, res_type, shape=None):
             (UInt, Float): arith_d.UIToFPOp,
             (Float, Int): arith_d.FPToSIOp,
             (Float, UInt): arith_d.FPToUIOp,
-            # FP to Index is not supported in MLIR
-            # (Float, Index): RuntimeError,
-            # (Index, Float): RuntimeError,
```
chhzh123, line 372: Can you also add float<->index testing?

chhzh123 (commented):
Mostly LGTM

zzzDavid:
Thanks, I fixed the init value issue in the test case, and added two test cases for float <-> index casting

chhzh123 (approved):
Thanks!

PR #258, opened 2024-11-13
[dataflow] Add instruction dispatching by daisy-chain for unified systolic array

AdrianLiu00:
## Description ##
This PR adds instruction dispatching for the unified systolic array. Instructions are decoded in PE00 and then dispatched to the other PEs in the array using a daisy-chain approach to alleviate fan-out.

### Examples ###
An example is provided within. The control signals first propagate across PEs in the first row and then spread down each column.

## Checklist ##

- [x] Code is well-documented


tests/dataflow/test_unified_systolic.py
```diff
@@ -6,112 +6,151 @@
...
-    fifo_C = df.array(df.pipe(dtype=int32, shape=(), depth=4), shape=(P0, P1))
+    # interconnect
+    fifo_R = df.array(df.pipe(dtype=int32, shape=(), depth=16), shape=(P0, P1 - 1))
+    fifo_C = df.array(df.pipe(dtype=int32, shape=(), depth=16), shape=(P0 - 1, P1))
+    inst_broad = df.array(df.pipe(dtype=bool, shape=(), depth=4), shape=(P1 - 1,))
+    inst_chain = df.array(df.pipe(dtype=bool, shape=(), depth=4), shape=(P0 - 1, P1))
 
     @df.kernel(mapping=[P0, P1])
-    def gemm(A: int32[M, K], B: int32[K, N], flowtag: bool, C: int32[M, N]):
+    def gemm(A: int32[M, K], B: int32[K, N], inst: bool[1], C: int32[M, N]):
```
chhzh123: Any reasons changing back to `bool[1]`?

tests/dataflow/test_unified_systolic.py
```diff
@@ -6,112 +6,151 @@
...
-        local_S: int32
         i, j = df.get_pid()
 
+        # --------------------------------------------------------
+        # Decode and Dispatch
+
+        with allo.meta_if(i == 0 and j == 0):
+            tag: bool = inst[0]
+            inst_broad[j].put(tag)
+            inst_chain[i, j].put(tag)
```
chhzh123, line 34: is this needed for [0,0]?

tests/dataflow/test_unified_systolic.py
```diff
@@ -6,112 +6,151 @@
...
+
+        with allo.meta_else():
+            with allo.meta_if(i == 0):
+                flowtag: bool = inst_broad[j - 1].get()
+            with allo.meta_else():
+                flowtag: bool = inst_chain[i - 1, j].get()
+
+            with allo.meta_if(i == 0 and j != P1 - 1):
+                inst_broad[j].put(flowtag)
```
chhzh123, line 43: It seems there will be both read and write in the same kernel?

tests/dataflow/test_unified_systolic.py
```diff
@@ -6,112 +6,151 @@
...
     if hls.is_available("vitis_hls"):
-        mod(A_flat, B_flat, flowtag, C_flat)
+        mod = df.build(top)
+        mod(A_flat, B_flat, insts, C_flat)
         print(C_flat)
-        C_truth = np.dot(A, B).flatten()
-        print(C_truth)
-        np.testing.assert_allclose(C_flat, C_truth, atol=1e-5)
-        print("Passed!")
+        C_tru = np.dot(A, B).flatten()
```
chhzh123: `C_truth`

tests/dataflow/test_unified_systolic.py
```diff
@@ -6,112 +6,151 @@
...
-    flowtag = True
-    mod = df.build(top)
 
+    flowtag: bool = False
+    insts = [flowtag]
+    if hls.is_available("vitis_hls"):
+        mod = df.build(top)
+        mod(A_flat, B_flat, insts, C_flat)
+        print(C_flat)
+        C_tru = np.dot(A, B).flatten()
```
chhzh123: `C_truth`

tests/dataflow/test_unified_systolic.py
```diff
@@ -6,112 +6,151 @@
 import allo.dataflow as df
 import allo.backend.hls as hls
 import numpy as np
+import pytest
```
chhzh123: I would suggest not using pytest for dataflow cases for now

chhzh123 (commented):
Some minor issues

chhzh123:
Also, it would be better to create a new branch for a PR next time. Otherwise, the commit history will become longer and longer

tests/dataflow/test_unified_systolic.py
AdrianLiu00: Just wanna show some decoding later. Like extracting the first elements in instructions as the configuration signal. I can change it back to the original single bool arguments if you think it's better.

tests/dataflow/test_unified_systolic.py
AdrianLiu00: Yeah, but the read and write operation are on different fifo, so I think that's not a problem.

tests/dataflow/test_unified_systolic.py
AdrianLiu00: Sure, I'll fix it and also aforementioned naming issues.

tests/dataflow/test_unified_systolic.py
chhzh123: Then how about creating a test case calculating two consecutive GEMM, one using output-stationary and the other using weight-stationary

tests/dataflow/test_unified_systolic.py
chhzh123: There're no warnings during csim?

AdrianLiu00: No, the csim is purely clean with no relevant warning.

tests/dataflow/test_unified_systolic.py
AdrianLiu00: I think this might be a higher dimensional test, and we could add that in later PR. I'll let it just be a single bool argument here.

AdrianLiu00:
You're right. I'll notice that next time.

tests/dataflow/test_unified_systolic.py
chhzh123: What do you mean by higher dimensional test. Can't we pass in an instruction array with size 2?

AdrianLiu00: It depends on where you'd like to make the decoding happen. We can talk about that in details offline later.

AdrianLiu00:
The minor issues are fixed. And I also pushed this design on board and passed the functionality test. You can review it again.

chhzh123 (approved):
LGTM. Thx!

PR #264, opened 2024-11-24
[Dataflow][Example] Packed systolic array for GEMM

jiahanxie353:
## Description ##
This PR tries to add a test case for packed systolic using the Dataflow module

### Proposed Solutions ###
I have added a test case called  `tests/dataflow/test_packed_systolic.py`; but I'm still trying to figure out the indexing problems on [this line](https://github.com/cornell-zhang/allo/blob/9abe4253ac2b7f3046e27bc15604af2544a264b5/tests/dataflow/test_packed_systolic.py#L61), which raises:
```c
  File "allo/allo/ir/builder.py", line 1232, in build_Subscript
    return ASTTransformer.build_bit_operation(ctx, node, val=val, idx=idx)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "allo/allo/ir/builder.py", line 1152, in build_bit_operation
    value = build_stmt(ctx, node.value)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "allo/allo/ir/builder.py", line 73, in __call__
    res = method(ctx, node, **kwargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "allo/allo/ir/builder.py", line 1228, in build_Subscript
    return ASTTransformer.build_memory_access(ctx, node, val=val, idx=idx)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "allo/allo/ir/builder.py", line 1117, in build_memory_access
    val.results[idx], value.result, ivs, affine_attr, ip=ctx.get_ip()
    ^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'results'
```

### Examples ###
Example is shown in the `tests/dataflow/test_packed_systolic.py`. Is there any obvious error in the example?

## Checklist ##

- [x] Code is well-documented

### DIFF
--- a/tests/dataflow/test_packed_systolic.py
+++ b/tests/dataflow/test_packed_systolic.py
@@ -1,4 +1,6 @@
-# test_packed_systolic.py
+# Copyright Allo authors. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
 import allo
 from allo.ir.types import int8, int16, int32
 import allo.dataflow as df
--- a/tests/dataflow/test_packed_systolic.py
+++ b/tests/dataflow/test_packed_systolic.py
@@ -1,0 +1,97 @@
+# test_packed_systolic.py
+import allo
+from allo.ir.types import int8, int16, int32
+import allo.dataflow as df
+import allo.backend.hls as hls
+import numpy as np
+
+L, D = 2, 2
+M, N, K = L, 1*D, D
+PP = 2
+P0, P1 = M // PP + 2, N + 2
+
+if PP == 2:
+    np_type = np.int16
+    allo_type = int16
+else:
+    raise ValueError(f"Unsupported packing factor: {PP}")
+
+@df.region()
+def top():
+    fifo_A = df.array(df.pipe(dtype=allo_type, shape=(), depth=4), shape=(P0, P1))
+    fifo_B = df.array(df.pipe(dtype=allo_type, shape=(), depth=4), shape=(P0, P1))
+
+    @df.kernel(mapping=[P0, P1])
+    def gemm(
+        X_packed: allo_type[L // PP, D],
+        W_packed: allo_type[D, 1 * D // PP],
+        Z_packed: allo_type[L // PP, 1 * D]
+    ):
+        i, j = df.get_pid()
+        # Peripheral kernels
+        with allo.meta_if(i in {0, M + 1} and j in {0, N + 1}):
+            pass
+        with allo.meta_elif(j == 0):
+            # i > 0
+            for k in range(K // PP):
+                fifo_A[i, j + 1].put(X_packed[i - 1, k])
+        with allo.meta_elif(i == 0):
+            # j > 0
+            for k in range(K // PP):
+                fifo_B[i + 1, j].put(W_packed[k, (j - 1)])
+        # drain
+        with allo.meta_elif(i == M + 1 and j > 0):
+            for k in range(K // PP):
+                b: allo_type = fifo_B[i, j].get()
+        with allo.meta_elif(j == N + 1 and i > 0):
+            for k in range(K // PP):
+                a: allo_type = fifo_A[i, j].get()
+        # main body
+        with allo.meta_else():
+            c: allo_type = 0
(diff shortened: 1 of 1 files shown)


chhzh123:
@jiahanxie353 Thanks for contributing! Please follow this [guideline](https://github.com/cornell-zhang/allo) to format your code.

jiahanxie353:
> @jiahanxie353 Thanks for contributing! Please follow this [guideline](https://github.com/cornell-zhang/allo) to format your code.

thanks for getting back @chhzh123 !
can you maybe send another link to the guideline? the one you sent was the homepage of allo

chhzh123:
https://cornell-zhang.github.io/allo/developer/index.html#id1

jiahanxie353:
> https://cornell-zhang.github.io/allo/developer/index.html#id1

thanks! I believe I have passed all the code-format testing

any advice on fixing the issue? I wonder why I can't index into an `int16[][]` type like this: `Z_packed[i - 1, j - 1][k * 8 : (k + 1) * 8]`

chhzh123:
Maybe you can try separate the memory access -- first load `Z_packed[i - 1, j - 1]`, increment the bits of [k * 8 : (k + 1) * 8], and then store back the new value. You can check out a data packing example [here](https://github.com/cornell-zhang/allo/blob/main/tests/dataflow/test_daisy_chain_gemm.py#L27-L35).

tests/dataflow/test_packed_systolic.py
```diff
@@ -0,0 +1,92 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+import allo
+from allo.ir.types import int8, int16, int32
+import allo.dataflow as df
+import allo.backend.hls as hls
+import numpy as np
+
+L, D = 2, 2
+M, N, K = L, 1 * D, D
```
chhzh123: What is `1` here? Maybe create a constant variable?

tests/dataflow/test_packed_systolic.py
```diff
@@ -0,0 +1,92 @@
...
+PP = 2
+P0, P1 = M // PP + 2, N + 2
+
+if PP == 2:
+    np_type = np.int16
+    allo_type = int16
```
chhzh123, line 16: Can you also add PP=4, 8 cases?

tests/dataflow/test_packed_systolic.py
jiahanxie353: I removed `L, D` and only kept `M, N, K` for clearance. Please let me know if you prefer to keep `L, D`

tests/dataflow/test_packed_systolic.py
jiahanxie353: Sure, I added `PP=4`.
Now I'm hard-coding each basic element size is `int8`, which forces `allo_type == int8 * PP`. I will make it parametric and add different `PP`-`allo_type` combinations

tests/dataflow/test_packed_systolic.py
```diff
@@ -0,0 +1,94 @@
...
+        Z_packed: allo_type[M // PP, N],
+    ):
+        i, j = df.get_pid()
+        # Peripheral kernels
+        with allo.meta_if(i in {0, M // PP + 1} and j in {0, N + 1}):
+            pass
+        with allo.meta_elif(j == 0):
+            # i > 0
+            for k in range(K):
+                fifo_A[i, j + 1].put(X_packed[(i - 1) * PP, k])
```
jiahanxie353, line 42: This line looks suspicious.
`X_packed` has type `allo_type[M, K // PP]`. And for example, when `M, N, K = 4, 4, 2`, `X_packed` type is `allo_type[4, 1]`. And if we have a for-loop: `for k in range(2)`. Suppose `k = 1`, not only it didn't go out of bound when accessing `X_packed` array, the results are all correct...

maybe we should look into this...

PR #270, opened 2024-11-29
[dataflow] Add multi C/output caches for daisy-chain SA example

AdrianLiu00:
## Description ##
This PR enhanced previous daisy-chain SA by adding multi C/output caches.
image

### Examples ###
See the example code inside.

## Checklist ##

- [X] PR's title starts with a category (e.g. [Bugfix], [IR], [Builder], etc)
- [X] Changes are complete (i.e. I finished coding on this PR)
- [X] All changes have test coverage (It would be better to provide ~2 different test cases to test the robustness of your code)
- [X] Code is well-documented

### DIFF
--- a/tests/dataflow/test_daisy_chain_gemm.py
+++ b/tests/dataflow/test_daisy_chain_gemm.py
@@ -8,58 +8,86 @@ import allo.backend.hls as hls
 import numpy as np
 
 M, N, K = 4, 4, 4
-P0, P1 = M + 1, N + 1
+P0, P1 = M + 2, N + 2
 
 
 @df.region()
 def top():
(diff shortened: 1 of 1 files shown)


chhzh123:
I remember you said `int8` types may lead to incorrect results. Can you be specific about the reasons?

AdrianLiu00:
I actually haven't figured out why. I don't think it's an overflow problem, it might be type related. But the problem is easy to reproduce, just replace all the types here with int8.

chhzh123:
Indeed, the int8 type failed with hardware emulation but passed C simulation. I'll merge this PR first and figure out the reason later.

PR #281, opened 2024-12-19
[dataflow] Enable csyn for previous design under new use-def chain

AdrianLiu00:
## Description ##
This PR enables the csyn for previous systolic design and also improves test session.

### Problems ###
Previous design was hard to achieve csyn directly from high-level programming, because the original use-def chain failed to search and mark every target partitioned array. We used to have to add the primitives manually in hls code.

### Proposed Solutions ###
The new use-def chain makes it easy to add partition primitives in the high-level program.

## Checklist ##

- [x] Code is well-documented

### DIFF
--- a/allo/backend/hls.py
+++ b/allo/backend/hls.py
@@ -168,7 +168,7 @@ class HLSModule:
         ext_libs=None,
         configs=None,
         func_args=None,
-        wrapping=None,
+        wrapping=True,
     ):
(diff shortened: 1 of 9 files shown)


AdrianLiu00:
By the way, I'll try to equip the unified systolic array with multi-cache in the next couple days. Then I will maybe sort out and categorize all the existing systolic designs in the dataflow directory to make it well-organized.

PR #292, opened 2025-02-03
[Simulator][Dataflow] Support parallel simulation for kernel with independent PEs

Oh-Afterglow:
## Description ##
This PR enables parallel simulation for kernels with independent PEs, i.e. no FIFO communications between PEs. It's provided as a backend `LLVMOMPModule` including 2 parts:
- A transformation pass that injects operations of the OpenMP dialect (`omp`) into kernel's MLIR code, which separates PE function calls into different `omp.section`s and enables parallelism among different PEs.
- A lowering pipeline that lowers the transformed MLIR code into the simulator executable for CPU (a mixture of `llvm` and `omp` dialect operations able to run on the MLIR execution engine).

### Examples ###
The test case of tiled GEMM is provided.

## Checklist ##

- [x] Code is well-documented


chhzh123 (commented):
Thanks for submitting a PR. Since we are building a simulator which is supposed to be different from the original LLVM backend, it'd be better to separate the implementation from the original one. For example, you can create a new file called `simulator.py` under `allo/backend` and specify the target as `"simulator"` instead of overwriting `"llvm"`.

Oh-Afterglow:
I've restructured the files as requested.

PR #308, opened 2025-02-28
[Simulator][Dataflow] Add implementation for streams (FIFOs) between threads

Oh-Afterglow:
## Description ##
This pull request adds an implementation of the Allo's stream type, which is the type of FIFOs between PEs, to the dataflow simulator backend, thus enables the multi-threaded simulation of kernels with PEs interchanging data.

The implementation mainly includes:
1. Lowering Allo stream type variables to circular queues being shared memory between 2 PE threads
2. Preserving memory consistency as well as maximum parallelism with lock-free queue implementation
3. Extending the lowering of Allo's struct type to support structs being function arguments

### Examples ###
The following testcases are each supplemented with a test of this implementation:
- tests/dataflow/test_1D_systolic.py
- tests/dataflow/test_producer_consumer.py
- tests/dataflow/test_smith_waterman_systolic.py
- tests/dataflow/test_systolic.py
- tests/dataflow/test_tiled_systolic.py
- tests/dataflow/test_weight_stationary_gemm.py

(Some test cases have had the simulator passed but are not modified due to their code structure.)

## Checklist ##

- [x] Code is well-documented


chhzh123:
Can you provide an example of a generated MLIR-OMP module in this PR description?

chhzh123:
Also mentioning the limitation of this PR would be great

chhzh123:
Since our CI server only has 4 cores, we need to set the limit of the number of threads. Otherwise, the CI got stuck in infinite loops. Can you try to reduce the number of PEs to <= 4 when testing the simulation?

PR #321, opened 2025-03-19
[Dataflow] DFG Implementation

he-andy:
## Description ##
Parses a MLIR module into a dataflow graph for autoscheduler analysis (https://github.com/cornell-zhang/allo/discussions/297)

### Proposed Solutions ###
MLIR module can now be parsed into a dataflow graph. Each node in the graph represents an affine kernel and is annotated with information regarding the access pattern, achievable II, relative first/last read/write times for each possible loop permutation of the affine kernel. An approximate DSP factor for the affine kernel is also computed. This information will be used to build the performance model described in stream-hls. 

DFG implementation also includes printing to dotgraph format for easy debugging. 
### Examples ###
The 3-Matmul example 
```python
def three_mm(A: int32[8, 8], B: int32[8, 8], C: int32[8, 8], D: int32[8, 8]) -> int32[8, 8]:
        E: int32[8, 8] = matrix_multiply(A, B)
        F: int32[8, 8] = matrix_multiply(C, D)
        return matrix_multiply(E, F)
```

will produce the dataflow graph with this structure 

image

## Checklist ##

- [x] Code is well-documented

### DIFF
--- a/allo/autoscheduler/dfg.py
+++ b/allo/autoscheduler/dfg.py
@@ -1,0 +1,496 @@
+# Copyright Allo authors. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+from collections import defaultdict
+import enum
+import sys
+import itertools
+from allo._mlir.dialects import (
+    func as func_d,
+    affine as affine_d,
+    memref as memref_d,
+)
+from allo._mlir.ir import WalkResult, Operation, AffineMap
+from allo.ir.types import MemRefType
+from .util import (
+    LoopInfo,
+    is_reduction_loop,
+    is_terminator,
+    get_minimal_access_pattern,
+    compute_loop_II
+)
+
+class DFGNodeType(enum.Enum):
+    AFFINE = 0
+    RET = 1
+    ALLOC = 2
+    CONST = 3
+
+class Edge:
+    """Represents a data dependency between nodes in the dataflow graph."""
+    def __init__(self, dst_id: int, value=None, src_op=None, dst_op=None):
+        self.id = dst_id        # Destination node ID
+        self.value = value      # The value/memref being communicated
+        self.src_op = src_op    # Source operation (e.g., store)
+        self.dst_op = dst_op    # Destination operation (e.g., load)
+    
+    def __repr__(self):
+        return f"Edge(id={self.id}, value={self.value.get_name()})"    
+
+class EdgeInfo:
+    def __init__(self, accessMap: AffineMap): 
+        self.accessMap = accessMap
+        self.first_element_time = 0
+        self.last_element_time = 0
+
+    def __repr__(self): 
+        return f"EdgeInfo(accessMap={self.accessMap}"
+
+class NodeInfo:
+    def __init__(self, permutation):
+        # Map permutation to edge info associated with that permutation
+        self.stores_map = {} 
+        self.loads_map = {}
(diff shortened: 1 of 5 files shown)


allo/autoscheduler/dfg.py
```diff
@@ -0,0 +1,554 @@
...
+    get_minimal_access_pattern,
+    compute_loop_II,
+)
+
+
+class DFGNodeType(enum.Enum):
+    AFFINE = 0
+    RET = 1
+    ALLOC = 2
+    CONST = 3
```
chhzh123, line 27: Only the loops and buffer information need to be recorded? What about the load/store ops?

allo/autoscheduler/dfg.py
```diff
@@ -0,0 +1,554 @@
...
+        self, block=None, dsp_factors=None, mem_r_ports=None, mem_w_ports=None
+    ):
+        self.block: Block = block
+        self.nodes: dict[int, Node] = {}  # Map from node ID to Node
+        self.in_edges = defaultdict(list)  # Map from node ID to incoming edges
+        self.out_edges = defaultdict(list)  # Map from node ID to outgoing edges
+        self.edges = []  # list of all edges
+        self.memref_edge_count = defaultdict(int)  # Count edges per memref
+        self.next_node_id = 0
+        self.dsp_factors = (
```
chhzh123, line 118: Any references?

allo/autoscheduler/util.py
```diff
@@ -199,3 +206,292 @@ def check_preprocess_ok(schedule: Schedule) -> bool:
...
+        ("step", int),
+        ("trip_count", int),
+    ],
+)
+
+
+def get_minimal_access_pattern(op: Operation, loop_info: list[LoopInfo]) -> AffineMap:
+    """
+    Analyzes an affine load or store operation and returns an AffineMap representing
+    minimal access function under the permutation provided in loop_info.
```
chhzh123, line 227: Can you define "minimal" here?

chhzh123 (approved):
This PR is pretty neat! The DFG visualization is also pretty useful for debugging! I only have some clarification questions.

allo/autoscheduler/dfg.py
he-andy: load/store ops not allowed in top level of a kernel, must be contained inside a for loop. technically this could be generalized to support load and stores at top level too, but at this point in the pass we would've already asserted that load/stores are not present as a top level op. 

This is to be consistent with the stream-hls definition of a perfect affine kernel, but I think stores/loads could also be considered a special case of affine kernels (single iter for loop). And we could definitely add this in the future to have more flexibility for kernel authors.

allo/autoscheduler/dfg.py
he-andy: I think this only needs to be a rough estimate, so I asked Claude to produce a reasonable guess for each operator. This can also be customized by the user if these estimates are not appropriate. Stream-HLS used a less complete version of this to estimate DSP usage, only considering addf and mulf to the total DSP count.  (https://github.com/UCLA-VAST/Stream-HLS/blob/0f7d915d238147548f7328dc9b0545258238024d/lib/Support/DFG.cpp#L1709).

allo/autoscheduler/util.py
he-andy: added in latest commit

PR #331, opened 2025-03-26
[Refactor] Change Polybench Kernels Visibility to Public 

he-andy:
## Description ##
The kernels provided in the `examples/polybench` folder can be used to write other test cases, but currently can only be accessed through a function that attaches scheduling primitives. 

### Proposed Solutions ###
All the kernels and their reference numpy reference implementations in `examples/polybench` are now publicly visible and can be imported into other test files. 

The numpy reference implementations now also return values in addition to modifying the tensors in place (previously some would only modify in place) for consistency. 
## Checklist ##

- [x] Code is well-documented


examples/__init__.py
chhzh123, line 1: Just confirming, this is not exposed to the Allo library, right? For example, if people want to use a kernel in Polybench, they will write sth like

```python
import examples.polybench.adi
```

if the path is added to the Python environment.

he-andy: ~I think it will be `allo.examples.` for someone who installs allo locally. Inside `allo/tests` it will be `import examples.polybench`~

Nvm I just checked, it will be `examples.polybench.adi`

chhzh123: but the current `examples` folder is not a subfolder of `allo`, so in your implementation, I don't think it can be imported using `allo.examples`. Also, it is not recommended to make it a submodule of Allo, as it may introduce cyclic dependency, so plz make sure it is not included under `allo`

he-andy: I've double checked by pip installing again and making a test file in another directory and it will not be included under `allo.examples.` and will be under`examples.` as desired

PR #356, opened 2025-05-01
[Backend][AIE] Remove redundant local buffer through buffer relaying in builder and support zero external kernel

EthanMeng324:
## Description ##
This PR eliminates redundant local buffers by enabling buffer relaying in the builder. Specifically, if the left-hand side of an assignment is an already defined tensor, the result of the right-hand side computation is written directly into it, avoiding the creation of a new buffer and the subsequent `memref.copy`. The corresponding case will be the assignment for `A[:]` or `A[:, :]`. To ensure correctness, the target tensor is zero-initialized beforehand. This PR also adds support for external vectorized zero-initialization kernels in MLIR-AIE. Additionally, it fixes the recursive lookup strategy in the `analyze_read_write_pattern` function.

### Examples ###
For this GEMM design,
```python
TyI, TyO = int16, int32
M, N, K = 128, 128, 128
P0, P1 = 4, 4

@df.region()
def top():
    @df.kernel(mapping=[P0, P1])
    def gemm(A: TyI[M, K] @ LyA, B: TyI[K, N] @ LyB, C: TyO[M, N] @ LyC):
        C[:, :] = allo.matmul(A, B)
```
the original generated mlir in each core is like this:
```mlir
%core_0_2_tile_comp_gemm_0_0 = aie.core(%tile_comp_gemm_0_0) {
  %global_c0 = arith.constant 0 : index
  %global_c1 = arith.constant 1 : index
  %c9223372036854775807 = arith.constant 9223372036854775807 : index
  scf.for %arg0 = %global_c0 to %c9223372036854775807 step %global_c1 {
    %fifo0 = aie.objectfifo.acquire @in_mem_A_0R(Consume, 1) : !aie.objectfifosubview<memref<32x128xi16>>
    %local0 = aie.objectfifo.subview.access %fifo0[0] : !aie.objectfifosubview<memref<32x128xi16>> -> memref<32x128xi16>
    %fifo1 = aie.objectfifo.acquire @in_mem_B_R0(Consume, 1) : !aie.objectfifosubview<memref<128x32xi16>>
    %local1 = aie.objectfifo.subview.access %fifo1[0] : !aie.objectfifosubview<memref<128x32xi16>> -> memref<128x32xi16>
    %fifo_out0 = aie.objectfifo.acquire @out_mem_C_00(Produce, 1) : !aie.objectfifosubview<memref<32x32xi32>>
[...]

PR #360, opened 2025-05-04
[Backend][AIE] Experimental mlir-aie codegen using MLIR builder

Fangtangtang:
## Description ##
This PR enables MLIR-AIE codegen in Allo by setting up a compatible environment and introducing a lowering method from Allo IR to MLIR-AIE dialects via MLIR context conversion.

### Problems ###
Previously, the environment setup led to conflicts when importing both `allo` and `mlir-aie` simultaneously. MLIR-AIE code generation in Allo relied on string matching and substitution. 

### Proposed Solutions ###
- Configure a compatible environment using the IRON library, `mlir-aie`, and `llvm-aie` compilers from the v1.0 release wheels.
- Manually construct codegen logic to lower Allo IR into MLIR-AIE dialects, migrating the IR from Allo’s MLIR context into the MLIR-AIE context.

### Examples ###
```python
Ty = int32
M = 1024

@df.region()
def top():
    @df.kernel(mapping=[1])
    def core(A: Ty[M], B: Ty[M]):
        B[:] = allo.add(A, 1)

A = np.random.randint(0, 100, M).astype(np.int32)
mod = df.build(top, target="aie-mlir")
B = np.zeros(M).astype(np.int32)
mod(A, B)
np.testing.assert_allclose(B, A + 1)
print("PASSED!")
```
Please refer to tests/dataflow/aie/test_vector_add.py for details.

## Checklist ##

- [x] Code is well-documented


chhzh123:
@Fangtangtang Thanks for contributing! Could you follow [this guide](https://cornell-zhang.github.io/allo/developer/index.html#id1) to format your code?

chhzh123:
You can just add `pylint disable=import-error` at the top of the file

chhzh123:
Let me know when you finish the changes

Fangtangtang:
I’ve finished all the non-streaming cases except for the one using `bf16`.

chhzh123:
@EthanMeng324 Could you help review this PR first?

EthanMeng324:
> @EthanMeng324 Could you help review this PR first?

Sure.

allo/backend/experimental_aie/__init__.py
```diff
@@ -0,0 +1,236 @@
...
+            "func.func(convert-linalg-to-affine-loops),lower-affine",
+        ]
+        pipeline = f'builtin.module({",".join(passes)})'
+        with self.allo_module.context:
+            mlir_pass_manager.parse(pipeline).run(self.allo_module.operation)
+        top_func, core_func_groups, external_funcs = classify_aie_functions(
+            self.allo_module
+        )
+        # TODO: maybe use other ways to capture the relationship between DTensor, function group
+        inputs, outputs = self.collect_io(core_func_groups)
```
EthanMeng324: Can you follow the newest build flow as in #356 to reorder the build flow here?

allo/backend/experimental_aie/utils.py
```diff
@@ -0,0 +1,464 @@
...
+
+def inject_external_kernels(module: allo_ir.ir.Module) -> tuple[dict[str, bool], dict]:
+    """
+    Inject external kernels for compute cores.
+
+    Return:
+        - use_external_kernels: func_name -> use external
+        - injected_kernels: kernel name -> external code snippet
+        - include_src: source file inclusion in exteranl code
+    """
```
EthanMeng324: Can you add more comment here following the original format in `aie.py`, such as the detailed type and explanation for each argument and return?

allo/dataflow.py
```diff
@@ -312,6 +314,24 @@ def build(
...
+        # aie_mod = AIE_MLIRModule(
+        #     s.module,
+        #     s.top_func_name,
+        #     s.func_args,
+        #     project,
+        #     stream_info,
+        # )
+        # aie_mod.build()
+        # return aie_mod
+
```
EthanMeng324: Why is this commented out?

allo/backend/experimental_aie/README.md
```diff
@@ -0,0 +1,68 @@
...
+```
+
+**After:**
+
+```python
+def downgrade_ir_for_peano(llvmir):
+    llvmir = llvmir.replace("getelementptr inbounds nuw", "getelementptr inbounds")
+    llvmir = llvmir.replace("captures(none)", "")
+    return llvmir
+```
```
EthanMeng324: Is this a common problem to all users? If so, is there anyway to automate this procedure?

allo/backend/experimental_aie/memory.py
```diff
@@ -0,0 +1,124 @@
+# Copyright Allo authors. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from ...memory import Layout, DTensor
+
+
+class AIE_DTensor:
+    """
+    Distributed tensor used in experimental AIE
+    """
```
EthanMeng324: Is there any specific reason to create a new class instead of just using the original `Layout` and `DTensor`?

allo/backend/experimental_aie/mlir_codegen.py
```diff
@@ -0,0 +1,766 @@
...
+    """
+    Current constrains:
+        - use 4 mem-shim tile pairs for io
+        - each port is assigned to one dtensor tile
+
+    Return:
+        - tile_map: dtensor name -> a list of dma tiles
+        - mem_tile_num
+        - shim_tile_num
+    """
```
EthanMeng324: Can you preserve the comment of the function from the original `aie.py`, especially the `TODO`? Also, if the functionality remains unchanged or closed to the original function, you can just invoke the original function to avoid duplicate code.

tests/dataflow/aie/experimental/test_matrix_add.py
EthanMeng324, line 1: Since the implementation of the kernel remains the same, we don't need to create new tests just for a different backend. If all the test has passed locally, you can just change the target to `aie-mlir` for all the tests in `tests/dataflow/aie` folder. Or you can also check the environment to determine using `aie` or `aie-mlir` backend.

EthanMeng324 (commented):
Thanks for contributing! The code is very well-organized. Just some comments below.

allo/backend/experimental_aie/memory.py
Fangtangtang: Only a new class is created as a replacement for `DTensor`.
This is just a temporary workaround — I'm planning to add some features to `DTensor`, but I don't want the experimental version to interfere with the original code.

allo/backend/experimental_aie/README.md
Fangtangtang: I'm not sure if this is a common problem, but I've been running into it myself, and this patch works for me.
If I had to automate the procedure, I'd probably write a script to detect and rewrite part of the `pip install`-generated files — but that doesn't seem like a good solution to me.
I'm still working through some environment issues and hoping to resolve everything cleanly during the initial setup phase.

allo/backend/experimental_aie/memory.py
EthanMeng324: Ok, you can add some comments there then.

allo/backend/experimental_aie/README.md
EthanMeng324: Sure, then we can just leave it there and save for the next PR.

allo/dataflow.py
Fangtangtang: I originally assumed that only one of the two codegen versions — `aie` or `aie-mlir` — would eventually be kept. Since we're currently using the `aie` version, I commented out `aie-mlir` for now.😓

EthanMeng324: We can have both options for now, so we can use both for the test.

Fangtangtang: I'm a bit concerned about compatibility — `experimental_aie` needs to import both `allo` and `aie`, and as far as I remember, that isn't supported in the current environment. This could cause import errors or conflicts during testing.

Fangtangtang:
@EthanMeng324 Thanks for the feedback! I've made the changes accordingly — please take a look and let me know if it now meets your expectations.

EthanMeng324:
> @EthanMeng324 Thanks for the feedback! I've made the changes accordingly — please take a look and let me know if it now meets your expectations.

LGTM, thanks! @chhzh123

allo/backend/experimental_aie/README.md
chhzh123, line 1: Just use `experimental` as the folder name instead of `experimental`

allo/backend/experimental_aie/README.md
```diff
@@ -0,0 +1,268 @@
...
+
+Then, install Allo as usual:
+```bash
+git clone https://github.com/cornell-zhang/allo.git && cd allo
+python3 -m pip install -v -e .
+```
+
+### Commands Used
+
+Below are the exact commands I used to set up the environment:
```
chhzh123: Use second-person narration throughout this README

allo/backend/experimental_aie/memory.py
chhzh123, line 1: What's the difference between `AIE_DTensor` and `DTensor`?

allo/backend/experimental_aie/mlir_codegen.py
```diff
@@ -0,0 +1,831 @@
...
+                    get_element_allo_type(dtype),
+                )
+                self.shape = shape
+                self.dtype = dtype
+                self.is_tensor = len(shape) > 0
+        else:
+            raise ValueError(f"Invalid stream type {type_str}.")
+
+    def __str__(self):
+        return f"Stream (name={self.name}, depth={self.depth}, allo dtype={self.allo_elememt_type}, is tensor={self.is_tensor}, src={self.src}, dst={self.dst})"
```
chhzh123: `dtype`, `is_tensor`

chhzh123 (commented):
Thanks for migrating the AIE codegen. This PR looks pretty nice! Just some clarification questions listed below

allo/backend/experimental_aie/memory.py
Fangtangtang: `AIE_DTensor` is mainly an extension of `DTensor`. The key difference is that [`calculate_tensor_access`](https://github.com/cornell-zhang/allo/blob/6a31067d279bf0ba447915c6a991373c64cbf217/allo/backend/aie.py#L542) is implemented as a member function of `AIE_DTensor`.

My intention was just to make the use of `DTensor` here clearer.

Fangtangtang:
@chhzh123 I've merged `AIE_DTensor` and `DTensor` as discussed. Please let me know if anything else needs to be updated.

chhzh123 (approved):
LGTM. Thanks!

PR #367, opened 2025-05-27
[Backend][AIE][Profiling] Enable profiling AIE module and fix bugs

Fangtangtang:
## Description ##
This PR introduces a new feature to the `AIE_MLIRModule` class. The goal is to enable users to easily profile the performance of their module executions by measuring runtime over multiple iterations. 

### Problems ###
Previously, there was no built-in way to evaluate the performance of modules.
Previously, the `top_func_name` when building aie module was hardcoded as `top`.

### Proposed Solutions ###
- Extend host code construction to include profiling option.
- Added `profile`, `warmup`, and `num_iters` parameters to the `build` method.
  - When `profile=True`, the module performs:
    - A configurable number of warm-up runs (default: 20) to stabilize performance.
    - A configurable number of timed runs (default: 100) to gather average execution time.
    - Print average execution time to the console.

- Fix bug to enable different `top_func_name`.
### Examples ###
```python
import allo
from allo.ir.types import int16, int32, float32
import allo.dataflow as df
import numpy as np
from allo.memory import Layout

Ty = int16
M, N, K = 128, 128, 32
Pm, Pn, Pk = 4, 4, 1
Mt, Nt, Kt = M // Pm, N // Pn, K // Pk

LyA = Layout("S1S2")
LyB = Layout("S2S0")
LyC = Layout("S1S0")

@df.region()
def top1():
    @df.kernel(mapping=[Pk, Pm, Pn])
    def gemm(A: Ty[M, K] @ LyA, B: Ty[K, N] @ LyB, C: int32[M, N] @ LyC):
        C[:, :] = allo.matmul(A, B)

mod = df.build(
    top1,
    target="aie-mlir",
    profile=True,
    warmup=200,
    num_iters=1000,
)
A = np.random.randint(0, 32, (M, K)).astype(np.int16)
B = np.random.randint(0, 32, (K, N)).astype(np.int16)
C = np.zeros((M, N)).astype(np.int32)
tmp_C = np.zeros((M, N)).astype(np.int32)
mod(A, B, C)
```
Sample output:
```txt
Avg NPU execution time: 406.05us
```

## Checklist ##

- [x] Code is well-documented

### DIFF
--- a/allo/backend/experimental/README.md
+++ b/allo/backend/experimental/README.md
@@ -306,18 +306,24 @@ print("PASSED!")
 #### Profiling
 A new profiling feature has been added to help measure the performance of the module during execution. 
 
-To enable profiling, use the `do_profile` flag in the `__call__` method in the [`AIE_MLIRModule` class](__init__.py):
+To enable profiling, use the `do_profile` flag in the `build` method in [`dataflow.py`](../../dataflow.py):
 ```python
-def __call__(
-    self,
-    *args,
-    do_profile: bool = False,
-    warmup_iterations: int = 20,
-    test_iterations: int = 100,
+def build(
+    func,
+    target="vitis_hls",
+    mode="csim",
+    project="top.prj",
+    configs=None,
+    wrap_io=True,
+    opt_default=True,
+    enable_tensor=False,
+    do_profile=False,
+    warmup_iterations=20,
+    test_iterations=100,
 ):
 ```
 
-**Parameters:**
+**New Parameters:**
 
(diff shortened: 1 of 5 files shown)


Fangtangtang:
> I'd suggest passing the `do_profile` argument in the `build` function instead of the `__call__` function. Specially, we can use this signature:
> 
> ```python
> def build(self, profile: bool = False, num_iters: int = 100, warmup: int = 20)
> ```

Changes made as suggested. Please let me know if there’s more to do.

PR #368, opened 2025-06-04
[AIE] Support for User-Defined External Kernels

Fangtangtang:
## Description ##
This pull request adds support for user-defined external kernels for MLIR-AIE backend.

Currently, developers were limited to the set of [predefined external kernels](https://github.com/Xilinx/mlir-aie/tree/v1.0/aie_kernels) provided in the repository, which covered only a narrow range of operations. 
With this PR, users can now integrate their own custom AIE kernels written in C++ and exposed through extern "C" interfaces.

In addition, this PR also fixes a bug related to DTensor addressing patterns.
### Problems ###
Currently,  complex computations on AIE cores were implemented using a limited set of [external kernels provided in the mlir-aie repository](https://github.com/Xilinx/mlir-aie/tree/v1.0/aie_kernels), which covered only a narrow range of operations and leaves room for performance improvement. 

### Proposed Solutions ###
This PR introduces support for user-defined external kernels. It provides an interface for users to register custom external kernels and use them within Allo kernels.

### Examples ###
Implement an external kernel in `norm.cc`
```cpp
#include <aie_api/aie.hpp>
#include <stdint.h>
#include <stdio.h>
#include <stdlib.h>
#include <type_traits>

#define NOCPP

#define EPS 1e-6f // epsilon

template <typename T_in, typename T_out, const int SEQ_LEN, const int HIDDEN>
void rms_norm_single_batch(T_in *input_tensor, T_in *weight,
                           T_out *output_tensor) {
  constexpr int vec_factor = 16;
  using vec_t = aie::vector<T_in, vec_factor>;
  event0();
  for (int iter = 0; iter < SEQ_LEN; iter++) {
    T_in *__restrict input_ptr = input_tensor;
    T_in *__restrict weight_ptr = weight;
    T_out *__restrict output_ptr = output_tensor;
    float square_sum = 0.0f;
    const int F = HIDDEN / vec_factor;
    for (int i = 0; i < F; i++) {
      vec_t input_vec = aie::load_v<vec_factor>(input_ptr);
      input_ptr += vec_factor;
      vec_t square_vec = aie::mul(input_vec, input_vec);
      square_sum += aie::reduce_add(square_vec);
    }
    vec_t square_sum_vec =
        aie::broadcast<T_in, vec_factor>(square_sum / HIDDEN + EPS);
    vec_t rms = aie::invsqrt(square_sum_vec);
    input_ptr = input_tensor;
    for (int i = 0; i < F; i++) {
      vec_t input_vec = aie::load_v<vec_factor>(input_ptr);
      input_ptr += vec_factor;
      vec_t normed = aie::mul(input_vec, rms);
      vec_t weight_vec = aie::load_v<vec_factor>(weight_ptr);
      weight_ptr += vec_factor;
      vec_t result = aie::mul(normed, weight_vec);
      aie::store_v(output_ptr, result);
      output_ptr += vec_factor;
    }
    input_tensor += HIDDEN;
    output_tensor += HIDDEN;
  }
  event1();
}

// exposed via extern "C" interfaces
extern "C" {
void rms_norm(float A_in[4][512], float B_in[512], float C_out[4][512]) {
  rms_norm_single_batch<float, float, 4, 512>(&A_in[0][0], B_in, &C_out[0][0]);
}
}

```
Register with `ExternalModule` and use in an Allo kernel.
```python
    # register user-defined external kernel
    norm = ExternalModule(
        top="rms_norm", # Name of the top-level function defined with `extern "C"`
        impl_path="norm.cc", # Path to the user-provided source file that implements the external kernel
        input_idx=[0, 1], # Indices of input arguments in the argument list
        output_idx=[2], # Indices of output arguments in the argument list
    )

    Ty = float32
    M, N = seq_len, hidden_size

    @df.region()
    def top():
        @df.kernel(mapping=[1])
        def core(A: Ty[M, N] @ LyA, B: Ty[N] @ Ly, C: Ty[M, N] @ LyA):
            norm(A, B, C) # use registered kernel
```

## Checklist ##

- [x] Code is well-documented


chhzh123:
bugbot run

chhzh123:
BugBot run

chhzh123:
BugBot run

allo/backend/__init__.py
```diff
@@ -1,4 +1,7 @@
 # Copyright Allo authors. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
-
+import os
 from . import llvm, hls, ip, aie
+
+if os.getenv("USE_AIE_MLIR_BUILDER") == "1":
+    from . import experimental
```
chhzh123: I think this guard should be enforced during compilation? Does it incur any issues when importing `experimental`?

allo/backend/experimental/external_kernel.py
```diff
@@ -0,0 +1,59 @@
+# Copyright Allo authors. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+import re
+import os
+from pyparsing import Keyword, Literal, nestedExpr, originalTextFor
```
chhzh123, line 6: Is this an external Python library?

allo/ir/builder.py
```diff
@@ -64,6 +65,9 @@
 from ..utils import get_mlir_dtype_from_str
 from ..logging import print_error_message
 
+if os.getenv("USE_AIE_MLIR_BUILDER") == "1":
+    from ..backend.experimental.external_kernel import ExternalModule
```
chhzh123: Just to understand, what will happen if these guards are not added? It does not seem to be a good practice to guard it everywhere

tests/dataflow/aie/norm.cc
```diff
@@ -0,0 +1,58 @@
...
+    input_tensor += HIDDEN;
+    output_tensor += HIDDEN;
+  }
+  event1();
+}
+
+extern "C" {
+void layer_norm(float A_in[4][512], float B_in[512], float C_out[4][512]) {
+  rms_norm_single_batch<float, float, 4, 512>(&A_in[0][0], B_in, &C_out[0][0]);
+}
```
chhzh123: As you include the layernorm function, could you also add test cases for this?

chhzh123 (commented):
Thanks! Mostly LGTM

allo/backend/__init__.py
Fangtangtang: This will need to import aie, which causes an import error when running PR checks.

allo/backend/experimental/external_kernel.py
Fangtangtang: Yes

allo/ir/builder.py
Fangtangtang: An import error when running PR checks. Because aie cannot be imported in current environment.
image

tests/dataflow/aie/norm.cc
Fangtangtang: A corresponding test can be found in `test_norm.py`

tests/dataflow/aie/test_norm.py
Fangtangtang, line 1: Here is a test case for the norm function.

tests/dataflow/aie/norm.cc
chhzh123: Layernorm and RMS norm are not the same? We'd better distinguish them

chhzh123:
@Fangtangtang Can you add `mlir-aie` and `pyparsing` to [`requirements.txt`](https://github.com/cornell-zhang/allo/blob/main/requirements.txt) so it can be automatically installed in our CI environment?

chhzh123:
Actually I don't know whether it will have conflicts when we also have an `aie` file under our package
```
allo/backend/__init__.py:4: in <module>
    from . import llvm, hls, ip, aie, experimental
allo/backend/experimental/__init__.py:9: in <module>
    import aie.ir as aie_ir
```

Probably you may also need to set up the AIE environment under [this file](https://github.com/cornell-zhang/allo/blob/main/.github/workflows/config.yml) if needed

Fangtangtang:
> Actually I don't know whether it will have conflicts when we also have an `aie` file under our package
> 
> ```
> allo/backend/__init__.py:4: in <module>
>     from . import llvm, hls, ip, aie, experimental
> allo/backend/experimental/__init__.py:9: in <module>
>     import aie.ir as aie_ir
> ```
> 
> Probably you may also need to set up the AIE environment under [this file](https://github.com/cornell-zhang/allo/blob/main/.github/workflows/config.yml) if needed

I guess it's more likely an environment variable issue. I'll try

Fangtangtang:
> Actually I don't know whether it will have conflicts when we also have an `aie` file under our package
> 
> ```
> allo/backend/__init__.py:4: in <module>
>     from . import llvm, hls, ip, aie, experimental
> allo/backend/experimental/__init__.py:9: in <module>
>     import aie.ir as aie_ir
> ```
> 
> Probably you may also need to set up the AIE environment under [this file](https://github.com/cornell-zhang/allo/blob/main/.github/workflows/config.yml) if needed

Yeah, seems like the conflict is one of the issues. Is it okay to rename the `aie` file under our package? we cannot modify the mlir-aie wheels

Fangtangtang:
@chhzh123 I’ve completed the tests for RMSNorm and LayerNorm. I also updated the environment and temporarily renamed `allo/backend/aie.py` to `allo/backend/ai_engine.py.`
Please let me know if there’s anything else I should do.

PR #362, opened 2025-05-05
[Autoscheduler] Node parallel performance model

he-andy:
## Description ##
Implements the node parallel performance model described in the stream-hls paper (https://github.com/cornell-zhang/allo/discussions/297)

### Proposed Solutions ###
The previous MINLP formulation was adapted to use a symbolic expression for the first/last read/write times so that they could be expressed as a function of the loop tiling factor. The permutations are then fixed while the tiling factors are optimized for minimizing latency while respecting the DSP and tiling equality constraints. 
### Examples ###
This kernel 

```python
def simple() -> float32[64, 64]:
        A: float32[64, 64]
        B: float32[64, 64]
        for i in range(64):
            for j in range(64):
                A[i, j] = i + j

        for j in range(64):
            for i in range(64):
                B[i, j] = A[i, j] + 1

        return B
```

Produces the following MINLP, which, when solved, yields the (correct) solution of fully tiling each for loop. 

```lp
Minimize
  0 lw1 + 0 st2 + 0 fw2 + 0 z_and_0_0_1_0 + 0 z_and_0_1_1_0
   + 0 z_and_0_0_1_1 + 0 z_and_0_1_1_1 + 0 arrive_0_1_0_0
   + 0 arrive_0_1_1_0 + 0 arrive_0_1_0_1 + 0 arrive_0_1_1_1
   + max_last_write_time
Subject To
 pinned_0_1: b0_1 = 1
 pinned_1_0: b1_0 = 1
 DSP_budget: DSP_0 + DSP_1 <= 2
 permutation_constraint_0: b0_0 + b0_1 = 1
 permutation_constraint_1: b1_0 + b1_1 = 1
 st_root_0: st0 = 0
 R6: fw_term_0_0 = 0
 R7: fw_term_0_1 = 0
 fw_constr_0: - st0 + fw0 - fw_term_0_0 - fw_term_0_1 = 0
 R9: fw_term_1_0 = 0
 R10: fw_term_1_1 = 0
 fw_constr_1: - st1 + fw1 - fw_term_1_0 - fw_term_1_1 = 0
 st_plus_lr_constr_0_1: - 2 st1 - rlr_term_0_1_0 - rlr_term_0_1_1
   + st_plus_lr_0_1 = 0
 epilogue_0_1: epilogue_0_1 - epi_term_0_1_0 - epi_term_0_1_1 = 0
 lw_term_0_1: - depend_0_1 - epilogue_0_1 + lw_term_0_1 = 0
[...]

PR #371, opened 2025-06-20
[Autoscheduler] Node Parallel Schedule

he-andy:
## Description ##
This PR implements the schedule extraction from the node parallel MINLP PR (https://github.com/cornell-zhang/allo/pull/362) and also contains some bug fixes and improvements for the MINLP construction. Notably, all performance models can now be constructed from a generalized version of the MINLP.

### Proposed Solutions ###
Based on the MINLP solution, the appropriate loop split factors are computed and buffers are explicitly partitioned into an n-d array of FIFOs. For instance, if we have a buffer of 10x10 and some pseudocode that looks like this 

```python
for i in range(10):
    for j in range(10):
         read arr[i][j]
```

with loop tiling factors Ti = 2, Tj = 2, the transformation yields
```python
for i in range(10):
    for j in range(10):
         read arr[i%2][j%2][i//2][j//2]
```
The modulo and floor division are elided after the loop splitting process, and the last to dimensions of `arr` are converted to a stream. See example below. 

### Examples ###
source
```python
 def stageB(A: int32[10, 10]) -> int32[10, 10]:
            B: int32[10, 10]
            for i in range(10):
                for j in range(10):
                    B[i, j] = A[i, j] + 1
            return B
```
result
```c
void affine_kernel_0(
  hls::stream< int32_t > v14[2][2] /* v14[2][2][5][5] */,
  ap_int<33> v15,
  int32_t v16[10][10]
) {     // L23
  #pragma HLS stream variable=v14 depth=-1
  l_S_i_0_i_outer1: for (int i_outer1 = 0; i_outer1 < 5; i_outer1++) {  // L24
    l_S_i_0_j_outer1: for (int j_outer1 = 0; j_outer1 < 5; j_outer1++) {        // L25
    #pragma HLS pipeline II=1
      l_i_inner1: for (int i_inner1 = 0; i_inner1 < 2; i_inner1++) {    // L26
      #pragma HLS unroll
        l_j_inner1: for (int j_inner1 = 0; j_inner1 < 2; j_inner1++) {  // L27
        #pragma HLS unroll
          int32_t v21 = v14[i_inner1][j_inner1].read(); // v14[i_inner1][j_inner1][i_outer1][j_outer1]; // L28
          ap_int<33> v22 = v21; // L29
          ap_int<33> v23 = v22 + v15;   // L30
          int32_t v24 = v23;    // L31
          v16[(i_inner1 + (i_outer1 * 2))][(j_inner1 + (j_outer1 * 2))] = v24;  // L32
        }
      }
    }
  }
}
```

## Checklist ##

- [x] Code is well-documented

### DIFF
--- a/tests/autoscheduler/test_passes.py
+++ b/tests/autoscheduler/test_passes.py
@@ -58,7 +58,6 @@ def test_simple(debug_point, kind):
         mod = optimized_schedule.build()
         input = np.zeros((10, 10), dtype=np.int32)
         np.testing.assert_allclose(mod(input), expected)
-        print("pass!")
 
     elif is_available("vitis_hls"):
         mod = optimized_schedule.build(
@@ -302,5 +301,4 @@ def test_gesummv(debug_point, kind):
 
 
 if __name__ == "__main__":
-    # pytest.main([__file__])
-    test_simple(debug_point="loop_opts", kind="node")
+    pytest.main([__file__])
--- a/tests/autoscheduler/test_passes.py
+++ b/tests/autoscheduler/test_passes.py
@@ -13,7 +13,7 @@ from allo.backend.hls import is_available
 
 kinds = [
     "graph",
-    # "node",
+    "node",
     # "combined"
 ]
 
@@ -43,7 +43,7 @@ def test_simple(debug_point, kind):
         return B
 
     s = allo.customize(simple)
-    
+
     # Create AutoschedulerConfig with the specified parameters
     cfg = AutoschedulerConfig.builder().with_debug_point(debug_point).with_kind(kind)
     optimized_schedule = dataflow_optimization_pass(s, cfg)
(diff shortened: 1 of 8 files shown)


tests/autoscheduler/test_passes.py
```diff
@@ -277,4 +302,5 @@ def test_gesummv(debug_point, kind):
 
 
 if __name__ == "__main__":
-    pytest.main([__file__])
+    # pytest.main([__file__])
+    test_simple(debug_point="loop_opts", kind="node")
```
chhzh123: Recover this?

chhzh123 (commented):
Can you briefly talk about how you transform the buffers into FIFOs? Is it through the `UnresolvedFIFOPrimitive` interface?

he-andy:
> Can you briefly talk about how you transform the buffers into FIFOs? Is it through the `UnresolvedFIFOPrimitive` interface?

So I'm not to sure how the Schedule.to() function works... but my understanding is that sometimes? you need to specify the target function in which the buffer is used. However, there's some transformations that are simpler to do in the IR prior using some references from the dfg before those references are invalidated by the outlining step. So the unresolved fifo primitive carries the node number in the dfg which is then resolved to a regular fifo primitive (using a map containing node # -> function name produced in the outlining phase) before it is applied to the schedule. 

I'm still unclear if having the function name when using .to() is a real requirement so this could be redundant if that isn't the case.

chhzh123:
Indeed, the function name is somehow a placeholder. We directly change all the related buffers to FIFOs, no matter which function they exist in

PR #373, opened 2025-06-23
[AIE][Test] Add GPT-2 and core ops implementation

Fangtangtang:
## Description ##
This PR provides a GPT-2 implementation for aie backend. 
It also includes independent test cases for key components of the model, such as GEMM and attention score computation.

In addition, this PR fixes a potential bug in `ExternalModule` to avoid naming conflicts between user-defined external modules and external kernel libraries.

### Examples ###
GPT-2 implementation: `tests/dataflow/aie/gpt2/test_gpt2_block.py`

attention score: 
- independent test case: `tests/dataflow/aie/gpt2/test_attn_score.py`
- external kernel: `tests/dataflow/aie/gpt2/attn_score.cc`

GEMM: `tests/dataflow/aie/test_gemm.py`
## Checklist ##

- [ ] Code is well-documented


allo/backend/experimental/external_kernel.py
```diff
@@ -45,6 +45,7 @@ def __init__(
         assert self.filename.endswith(
             ".cc"
         ), f"Expected a .cc file, but got: {self.filename}"
+        self.filename = self.filename.removesuffix(".cc") + "_.cc"
```
chhzh123, line 49: Any reason for this?

tests/dataflow/aie/gpt2/attn_score.cc
```diff
@@ -0,0 +1,52 @@
...
+          static_cast<T_out>(sum * scale);
+      tensor_b_ptr += K;
+    }
+    tensor_a_ptr += K;
+  }
+}
+
+extern "C" {
+
+void partial_attn_score(float A_in[32][64], float B_in[32][64],
```
chhzh123: What is attention score? `transpose_matmul_with_scale` is much clearer

tests/dataflow/aie/test_gemm.py
chhzh123, line 1: Modify the file name to sth like `test_gemm_temporal_reduction.py`.

tests/dataflow/aie/gpt2/layer_norm.cc
chhzh123, line 1: These two `.cc` files should be inside the `allo` folder? We can construct a library folder for `aie`.

chhzh123 (commented):
LGTM. Just some file organization and naming issues.

allo/backend/experimental/external_kernel.py
Fangtangtang: I want to avoid naming conflict with the mile-aie kernels we are using as builtin library.

tests/dataflow/aie/gpt2/attn_score.cc
Fangtangtang: `attn_score = (Q @ K.T) / sqrt(d)`

`transpose_matmul_with_scale` is used to calculate this in this GPT2 implementation

tests/dataflow/aie/gpt2/layer_norm.cc
Fangtangtang: ok. But if it's meant to be used as a library, wouldn't it be better to use templates? The current customized external kernel mechanism doesn't support templates yet.

Since these two files are currently tailored specifically for gpt2 (especially parameter shapes), I've temporarily placed them under the `gpt2` folder.
I could move the two templates into a dedicated library folder, and then have a `.cc` file under gpt2 that simply includes the library files. 

What do you think?

Fangtangtang:
Hi, I've fixed some naming issues you mentioned. could you take another look and see if things look better now?

Also, I added an implementation of masked softmax, but it's currently a bit unstable. There’s a strange bug where the result is partially undefined if it’s only run once. However, if it's run multiple times, the final result is correct. It feels like some memory-related bug, and it might not be on our side.

tests/dataflow/aie/gpt2/layer_norm.cc
Fangtangtang, line 1: These `.cc` files are specific for the gpt2 block and depend on certain fixed hyperparameters.

Fangtangtang:
Just added external kernels for softmax and gelu. The numerical error is much smaller compared to all previous versions. 
These kernels work, but haven't been carefully optimized.

allo/backend/experimental/external_kernel.py
chhzh123: Adding a comment for this would be better

tests/dataflow/aie/gpt2/layer_norm.cc
chhzh123: Since we may want to reuse those cc implementations later, it is fine not to make them templates, but it is still better to put them under the allo library. You can put it under `allo/backend/experimental/kernels`

PR #376, opened 2025-07-03
[AIE][Backend] Enabling Large-Scale Computation on AIE Backends through Virtual Mapping 

Fangtangtang:
## Description ##
This PR implemented the proposed virtual mapping primitives `bundle` and `chain` to enable larger scale computation on AIE backend.
In addition, to improve the performance of the generated code, this PR introduces a set of optimizations. 

With these enhancements, the performance on certain GEMM cases is now close to the manually optimized [mlir-aie templates](https://github.com/Xilinx/mlir-aie/tree/v1.0/programming_examples/basic/matrix_multiplication/whole_array).

_Note_:
- The current interface for expressing virtual mapping primitives is primarily designed for internal use and prototyping. It is not yet suitable as a user interface.
- The current implementation of processing the primitives incurs significant compile-time overhead, which is actively being optimized.

### Problems ###
The current custom mapping splits tasks across multiple logical PEs, each mapped to different physical PEs. Due to the limited memory on each PE, the supported task size on each PE is small. Partitioning the computation along the spatial dimension can increase the computational capacity, but the limited number of physical PEs still imposes a constraint. 
To support even larger-scale computations, it is necessary to partition the workload along the temporal dimension as well.

Besides, simply applying virtual mapping primitives is not sufficient to achieve high performance. Without further optimization, the generated code may contain redundant buffers, suboptimal memory access patterns, and unvectorized operations.

### Proposed Solutions ###
Introducing virtual PEs, allowing users to assume an unlimited number of PEs. 
The compiler would then automatically search for feasible mappings from virtual PEs to the limited physical PEs.

Designed and implemented several optimization passes to elimiate redundant buffers, transform data layout at transfer time and vectorization.

### Examples ###
Please refer to testcases under tests/dataflow/aie for more examples. (those staring with test_mapping)

## Checklist ##

- [ ] Code is well-documented


allo/backend/experimental/external_kernel.py
```diff
@@ -7,6 +7,31 @@
 from ..ip import parse_cpp_function
 
 
+class ExternalModuleBase:
```
chhzh123, line 10: Are you going to use this for other external modules?

allo/backend/experimental/utils.py
```diff
@@ -1,25 +1,188 @@
...
+    F16Type,
+    F32Type,
+    BF16Type,
 )
 
+
+# ############################################################
+# Configurations
+# ############################################################
+class Config:
```
chhzh123, line 39: Use [`dataclass`](http://docs.python.org/3/library/dataclasses.html)

chhzh123: Also, do all the NPU devices have the same config as below?

allo/backend/experimental/utils.py
```diff
@@ -52,14 +215,54 @@
     "ui64": "unsigned long",
 }
 
+matmul_externel_kernel_config_map = {
```
chhzh123: Add comment for this

mlir/include/allo/Dialect/AlloOps.td
```diff
@@ -1029,4 +1029,32 @@ def StreamPutOp : Allo_Op<"stream_put"> {
...
 
+//===----------------------------------------------------------------------===//
+// Layout operations
+//===----------------------------------------------------------------------===//
+
+def TransformLayoutOp : Allo_Op<"transform_layout"> {
+  let summary = "Access a memref with a new layout";
+  let description = [{
+    Takes a memref and three constant integer arrays: offsets, sizes, and strides per dimension.
+    Returns a new memref with updated layout.
```
chhzh123: Probably provide several examples here so developers can know the actual meaning of those arguments

mlir/lib/Bindings/AlloModule.cpp
```diff
@@ -27,6 +27,7 @@
 #include "mlir/Pass/PassManager.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 #include "mlir/Transforms/Passes.h"
+#include <tuple>
```
chhzh123: Where do you use tuple inside this file?

mlir/lib/Transforms/CopyOnWrite.cpp
```diff
@@ -0,0 +1,134 @@
...
+#include "mlir/Dialect/Linalg/IR/Linalg.h"
+#include "mlir/Dialect/MemRef/IR/MemRef.h"
+
+using namespace mlir;
+using namespace allo;
+
+namespace mlir {
+namespace allo {
+
+void removeRedundentCopy(func::FuncOp &func) {
```
chhzh123: typo: redundant

allo/backend/experimental/external_kernel.py
Fangtangtang: What are other external modules? I thought builtin and customized has covered all of them.

I want to use `ExternalModuleBase` as a unified interface for both built-in external kernels (from the mlir-aie repo and our own library) and user-provided custom kernels with the `ExternalModule` interface. 
My initial goal is to treat all external kernels in the same way. Our compiler have a pass that replaces certain ops with built-in external kernels. Ideally, if users could specify some form of rewrite rule when registering their `ExternalModule`, we could apply their custom kernels just like the built-in ones. 
However, I haven't found a good way to implement this yet

allo/backend/experimental/utils.py
Fangtangtang: Yeah, I think so for now. This config also works fine on xDNA2. 
If there are different configs in the future, we can simply add a configuration interface.

mlir/lib/Bindings/AlloModule.cpp
Fangtangtang: Sorry, I used that for some simple tests, forgot to delete this one

PR #392, opened 2025-08-09
[AIE][Backend][Bugfix] fix bug caused by file-based data transfer between python interface and cpp

Fangtangtang:
## Description ##
This PR fixes a bug in the AIE backend where data was passed between Python and C++ via file read/write. 

### Problems ###
Previous implementation had bug in the value parsing process during file read/write.

For example, the computation error in int8 GEMM was actually caused by reading the data from file in the C++ program as `char`, which resulted in the values being interpreted as ASCII codes when converted to int8.

### Proposed Solutions ###
Use binary mode for file read/write.

## Checklist ##

- [ ] Code is well-documented


chhzh123 (approved):
LGTM, thanks!

PR #404, opened 2025-09-03
[Dataflow][Examples] FEATHER Implementation

Oh-Afterglow:
## Description ##
This PR implements an accelerator [_FEATHER_](http://arxiv.org/abs/2405.13170) with Allo dataflow model as a test case. _FEATHER_ includes a spacial array called NEST for local reduction, and a multi-stage reduction network called BIRRD enabling reorder-in-reduction. Custom instructions can be input to NEST to specify a desired way of reduction and reordering. This implementation performs a weight-stationary GEMM while rearranging the columns of the result.

2 parameters, Array Width (`AW`) and Array Height (`AH`), control the size of NEST. `AW` determines the size of BIRRD. In this test case, `AH` also determines the size of input matrices for simplicity. This test case has 3 preset size configurations where AW=4, 8, 16. A new instruction array is needed if `AW` is changed to other values. The values can be passed as command-line arguments `--AW` and `--AH`.

This kernel can be simulated with the current dataflow simulator.

## Checklist ##

- [x] Code is well-documented

### DIFF
--- a/tests/dataflow/test_FEATHER_loop.py
+++ b/tests/dataflow/test_FEATHER_loop.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 from math import log2
+import argparse, sys
 
 import allo
 from allo.ir.types import float32, int8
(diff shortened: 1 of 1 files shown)


tests/dataflow/test_FEATHER_loop.py
```diff
@@ -0,0 +1,267 @@
...
+    #     [
+    #         [PS, PS, PS, PS],
+    #         [PS, PS, PS, PS],
+    #         [AR, AR, AL, AL],
+    #         [SW, SW, SW, SW],
+    #         [SW, PS, PS, SW],
+    #         [PS, PS, PS, PS],
+    #     ],
+    #     dtype=np.int8,
+    # )
```
chhzh123: Can you provide command-line arguments for configuring the size of the array so these instructions don't need to be in comments?

Oh-Afterglow: Thanks for the suggestion. A new commit adding command line functionality is already submitted.

chhzh123 (approved):
LGTM. Thx!

PR #423, opened 2025-09-18
[Backend][AIE] Generate scalar code for VLIW processor from Python

chhzh123:
## Description ##
It is a tedious process to write C kernels for the AIE VLIW processors, which involves complicated code rewriting and optimization. This PR takes the first step by exposing the Python interface for users to define these "external functions". We reuse the Allo frontend parser, the MLIR builder, and the C codegen to translate a Python program into a C kernel.

Currently, it only supports scalar kernels with straightforward translation. Later on, we need to add vectorization support using the MLIR vector dialect.

### Examples ###
The proposed interface is shown below. Users can use `@vliw.kernel()` to decorate their Python kernel, which will be later translated into a C kernel.

```python
def test_simple_add():

    @vliw.kernel()
    def simple_add(a: float32[16], b: float32[16]) -> float32[16]:
        c: float32[16] = 0.0
        for i in allo.grid(16):
            c[i] = a[i] + b[i]
        return c

    # Use in actual AIE dataflow:
    @df.region()
    def my_aie_design():
        @df.kernel(mapping=[1])
        def core(
            A: float32[16],
            B: float32[16],
            C: float32[16],
        ):
            simple_add_ext(A, B, C)

    aie_module = df.build(my_aie_design, target="aie")
    A = np.random.rand(16).astype(np.float32)
    B = np.random.rand(16).astype(np.float32)
    C = np.zeros(16).astype(np.float32)
    aie_module(A, B, C)
    # verify the result
    np.testing.assert_allclose(C, A + B)
    print("✓ AIE design verification successful")
```

Example generated VLIW C code:
```c
/*
 * Generated VLIW C code for AMD AIE processors
 * Function: simple_add
 */
#define NOCPP
#include <aie_api/aie.hpp>
#include <stdint.h>
#include <stdio.h>
#include <stdlib.h>
#include <type_traits>

extern "C" {

void simple_add(
  float v0[16],
  float v1[16],
  float v2[16]
) {	// L2
  for (int v3 = 0; v3 < 16; v3++) {	// L5
    v2[v3] = (float)0.000000;	// L5
  }
  l_S_i_0_i: for (int i = 0; i < 16; i++) {	// L6
    float v5 = v0[i];	// L7
    float v6 = v1[i];	// L8
    float v7 = v5 + v6;	// L9
    v2[i] = v7;	// L10
  }
}


} // extern "C"
```

## Checklist ##

- [x] Code is well-documented

### DIFF
--- a/allo/autoscheduler/dfg.py
+++ b/allo/autoscheduler/dfg.py
@@ -1,5 +1,7 @@
 # Copyright Allo authors. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
+# pylint: disable=no-name-in-module
+
 from collections import defaultdict
 from collections.abc import Callable
 from dataclasses import dataclass
@@ -9,13 +11,13 @@ import itertools
 from typing import Union, Optional
 import gurobipy as gp
 from gurobipy import GRB
-from allo._mlir.dialects import (
+from .._mlir.dialects import (
     func as func_d,
     affine as affine_d,
     memref as memref_d,
 )
-from allo._mlir.ir import WalkResult, Operation, AffineMap, Block, AffineExpr
-from allo.ir.types import MemRefType
+from .._mlir.ir import WalkResult, Operation, AffineMap, Block, AffineExpr
+from ..ir.types import MemRefType
 from .util import (
     LoopInfo,
     is_reduction_loop,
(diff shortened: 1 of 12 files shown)


chhzh123:
This PR also changed all the absolute import to local import

PR #427, opened 2025-09-23
[IR][Builder][Backend] Update variable scoping in IR builder

Fangtangtang:
## Description ##
This PR addresses several issues related to **variable scoping** in the IR builder.

### Problems ###
The previous IR builder was not robust in handling variable scopes. For example:

```python
C_: Ty[M, N] = 0
with allo.meta_for(Pk) as i:
    i = xxx
    C_[:, :] += pipe[i].get()
C[:, :] = C_
```
In this case, `pipe[i]` could still reference the iterator `i` from `meta_for`, leading to incorrect behavior.

Another problematic case was:
```python
if xxx:
    pk = df.get_pid()
pipe[pk].put(allo.matmul(A, B))
```
This was incorrectly considered valid, even though `pk` was declared inside the conditional scope.

(We adopt **C++-style block scope**: variables declared inside an `if/for/while/...` block are only valid within that block. This aligns better with MLIR.)

### Proposed Solutions ###
- Modified `TypeInferer` and `ASTTransformer` (IR builder) to properly handle variable scopes.
- Adjusted test cases with “invalid” syntax that previously compiled but should have been rejected.

Example before (incorrect scoping):

```python
@df.kernel(mapping=[Pk, Pm, Pn])
def gemm(A: TyI[M, K] @ LyA, B: TyI[K, N] @ LyB, C: TyO[M, N] @ LyC):
    pk, pm, pn = df.get_pid()
    with allo.meta_if(pk > 0):
        C_in: TyO[Mt, Nt] = pipe[pk - 1, pm, pn].get()
    with allo.meta_else():
        C_in: TyO[Mt, Nt] = 0
    C_out: TyO[Mt, Nt] = allo.add(allo.matmul(A, B), C_in) # should be invalid
    with allo.meta_if(pk < Pk - 1):
        pipe[pk, pm, pn].put(C_out)
    with allo.meta_elif(pk == Pk - 1):
        C[:, :] = C_out
```

After fix (explicit declaration outside block, assignments inside):

```python
@df.kernel(mapping=[Pk, Pm, Pn])
def gemm(A: TyI[M, K] @ LyA, B: TyI[K, N] @ LyB, C: TyO[M, N] @ LyC):
    pk, pm, pn = df.get_pid()
    C_in: TyO[Mt, Nt]
    with allo.meta_if(pk > 0):
        C_in[:, :] = pipe[pk - 1, pm, pn].get()
    with allo.meta_else():
        C_in[:, :] = 0
    C_out: TyO[Mt, Nt] = allo.add(allo.matmul(A, B), C_in)
    with allo.meta_if(pk < Pk - 1):
        pipe[pk, pm, pn].put(C_out)
    with allo.meta_elif(pk == Pk - 1):
        C[:, :] = C_out
```

## Checklist ##

- [ ] Code is well-documented


Fangtangtang:
Sorry about the mix-up. the scope issue came up while I was addressing some other problems, so part of the code got mixed. 
I’ve restored the AIE backend part.

chhzh123 (commented):
Can you provide additional test cases not limited to the AIE backend then? You can put them under `test_builder`

tests/test_builder.py
```diff
@@ -812,5 +812,79 @@ def kernel2(A: int32[10]) -> int32[10]:
...
+    def kernel4(n: int32) -> int32:
+        for i in range(n):
+            tmp: int32 = i
+        return tmp
+
+    with pytest.raises(SystemExit):
+        s = allo.customize(kernel4)
+
+    # case 5: inner loop shadows iterator from outer loop
+    def kernel5(n: int32) -> int32:
```
Fangtangtang, line 863: The scope of a `for` loop here is similar to C++ block scope, which differs from Python. The inner loop’s `i` shadows the outer loop’s `i`, but the `i` used in the outer loop body still refers to the outer iterator. 
Let me know if this behavior does not conform to our definition.

Fangtangtang:
I added some test cases, please check if they conform to our syntax definition.🫡

tests/dataflow/aie/test_cannon.py
```diff
@@ -45,14 +45,12 @@ def cannon(C: Ty[M, N] @ LyC):
             A_pipe[(pi - 1) % P, pj].put(A_out)
             B_pipe[pi, (pj - 1) % P].put(B_out)
 
-            with allo.meta_for(P - 1) as i:
+            for _ in range(P - 1):
```
chhzh123, line 48: Why changing this `meta_for` to `for`?

tests/dataflow/aie/test_mapping_gemm.py
```diff
@@ -91,10 +91,11 @@ def top():
         @df.kernel(mapping=[Pk, Pm, Pn])
         def gemm(A: Ty[M, K] @ LyA, B: Ty[K, N] @ LyB, C: Ty[M, N] @ LyC):
             pk, pm, pn = df.get_pid()
+            C_in: Ty[Mt, Nt]
             with allo.meta_if(pk > 0):
-                C_in: Ty[Mt, Nt] = pipe[pk - 1, pm, pn].get()
+                C_in[:, :] = pipe[pk - 1, pm, pn].get()
```
chhzh123, line 96: Can we just use `C_in = pipe[pk - 1, pm, pn].get()` here? Otherwise, it seems to become a copy

tests/test_builder.py
chhzh123: Yeah, I think this example is correct, but we probably want to document this later

allo/ir/visitor.py
```diff
@@ -325,6 +367,7 @@ def get_symbolic_expr(expr_node, mapping, var_map) -> str:
         - mapping:dict[str,str], the symbolic map (name in AST -> symbol)
         - var_map: name in AST -> value
     """
-    new_tree = ReplaceNames(mapping, var_map).visit(expr_node)
+    processor = ReplaceNames(mapping, var_map, variables)
```
chhzh123: Give a different name? `processor` is too overloaded

tests/dataflow/aie/test_cannon.py
Fangtangtang: `meta_for` is unrolled by default. 
Since `i` is not used, using `for` should be a better choice to reduce code size.

tests/dataflow/aie/test_mapping_gemm.py
Fangtangtang: I tried, that would lead to `Variable `C_in` has already been defined, please use a different name`.

Fangtangtang: And it should be a copy, right? 
if it’s defined as a variable inside the branch and initialized with `get()`, then according to scoping rules the variable would be destroyed once we exit the branch.

allo/ir/visitor.py
Fangtangtang: got it. I'll rename to `node_transformer`

tests/dataflow/aie/test_cannon.py
chhzh123: ok, this is a pure temporal loop, and all the stream indices are static

chhzh123 (approved):
LGTM, thx

Fangtangtang:
I noticed that many old test cases have incorrect scoping. The compiler used to handle scoping incorrectly, which accidentally led to the “correct” results. 
After this PR is merged, some of those tests will fail to compile. We’ll need to update them.

---
some of the buggy test cases：
- tests/dataflow/test_multi_cache_gemm.py
- tests/dataflow/test_unified_systolic.py
- tests/dataflow/test_daisy_chain_gemm.py
- tests/dataflow/test_unified_systolic.py

chhzh123:
I see. Please update them in a separate PR. We may need to enable dataflow simulation for our CI

PR #430, opened 2025-09-28
[Feature][Dataflow][AIE] Visualize virtual computation graph

Fangtangtang:
## Description ##
This PR introduces a new feature to visualize the virtual mapping graph.  

### Problems ###
Currently, there is no straightforward way to visualize the virtual mapping graph.  
This makes it difficult for users to identify the available virtual mapping primitives, understand how to apply them, and verify the mapping relationships for correctness, especially when debugging or analyzing complex cases.

### Proposed Solutions ###
Implemented a `dump` function in `ComputationGraph` to enable visualization of the virtual mapping graph.  
After building the dataflow module, users can find the corresponding PDF file in the build project directory.  

### Examples ###
**Example input program**:  `tests/dataflow/aie/test_cannon.py`
After building the dataflow module, the user can find the visualization result as a PDF file named "virtual_graph.pdf".

cannon

**Example input program**:  `_test_split_k_gemm_2x2x4` in `tests/dataflow/aie/test_mapping_gemm.py`
After building the dataflow module, the user can find the visualization result for the original virtual mapping graph as a PDF file named "virtual_graph.pdf".
before
And the visualization result for the virtual mapping graph after applying all the mapping primitives as a PDF file named "after_mapping.pdf".

after

## Checklist ##

- [ ] Code is well-documented

### DIFF
--- a/allo/backend/aie/README.md
+++ b/allo/backend/aie/README.md
@@ -353,6 +353,32 @@ ENABLE_AGGRESSIVE_PORT_UTILIZATION_PATCH=1
 ```
 This enables an aggressive strategy that tries to fully utilize available DMA ports.
 
+#### `DEBUG`
+
+This variable controls whether virtual mapping graphs are dumped.
+
+You can use the virtual mapping graph to identify the available virtual mapping primitives, understand how to apply them and verify if the mapping behaves as expected. 
+
+If you want to dump the virtual mapping graphs for debugging, you can set:
+```
+DEBUG=1
+```
+After building the dataflow module, you can find the visualization result as a PDF file in the build project directory.
+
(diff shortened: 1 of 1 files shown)


Fangtangtang:
The largest example we have is $2^{15}$, but there’s no upper bound. 
It definitely makes sense to only enable dumping in debug mode.

Do you think we should continue adding parameters to [the build function](https://github.com/cornell-zhang/allo/blob/main/allo/dataflow.py#L413) for this? It already has too many AIE-specific parameters.😅

chhzh123:
We can use `**kwargs` later, but the debug mode can be set as an environment variable

Issue #425, opened 2025-09-21
[BUG][AIE] Allow slice-based access for temporary buffer tensors (e.g., `buffer[i,:,:]`)


Fangtangtang:
**Describe the bug**
While implementing the `allreduce` pattern, I wanted to first experiment with an explicit `gather` and see how it can be lowered to mlir-aie. (This relates to our discussion in #406 about how the compiler internally constructs the virtual graph for `allreduce`, I'm trying to describe the pattern explicitly.)

This pattern requires using a loop to get multiple streams and store the values into a local buffer. However, such usage is currently rejected as illegal.

For example, in the following kernel, we use `buffer` as a 3D temporary tensor to store `Pk` different `(M, N)` slices, where each `buffer[i,:,:]` is filled from a pipe output and later accumulated into `C_`.
```python
buffer: Ty[Pk, M, N]
with allo.meta_for(Pk) as i:
    buffer[i,:,:] = pipe[i].get()
for i in range(Pk):
    C_[:, :] += buffer[i]
```

Currently this usage (`buffer[i,:,:]`) is rejected as illegal, even though it is a natural way to represent storing/reading per-iteration results in a temporary buffer.

**To Reproduce**

```python
Ty = int16
M, N, K = 32, 32, 128
Pk = 4

LyA = Layout("RS0")
LyB = Layout("S0R")

@df.region()
def top():
    pipe = df.array(df.pipe(dtype=Ty, shape=(M, N), depth=2), shape=(Pk,))

    @df.kernel(mapping=[Pk])
    def partial_gemm(A: Ty[M, K] @ LyA, B: Ty[K, N] @ LyB):
        pk = df.get_pid()
        pipe[pk].put(allo.matmul(A, B))

    @df.kernel(mapping=[1])
    def acc(C: Ty[M, N]):
        C_: Ty[M, N] = 0
        buffer: Ty[Pk, M, N]
        with allo.meta_for(Pk) as i:
            buffer[i,:,:] = pipe[i].get()
        for i in range(Pk):
            C_[:, :] += buffer[i] 
        C[:, :] = C_
```

**Expected behavior**
I expected this pattern to be accepted, with `buffer[i,:,:]` serving as a legal slice-based indexing into a temporary tensor.

Fangtangtang:
I think the root cause is that dynamic offset/size/stride slices are not supported yet. In addition, rank-reducing slices do not adjust the result data type, which leads to type mismatches in subsequent uses.

[...]

Issue #406, opened 2025-09-08
[Feature][AIE] Include `allreduce` operation


Fangtangtang:
The current large scale GEMM implementation is transformed from a ping-pong GEMM kernel. 
A better way to implement the same algorithm is to introduce an `allreduce` operation and construct streams (edges in the virtual computation graph) internally.

Possible interface and usage
```python
LyA = Layout("S1S2")
LyB = Layout("S2S0")
LyC = Layout("S1S0")

@df.region()
def top():
    @df.kernel(mapping=[Pk, Pm, Pn])
    def gemm(A: Ty[M, K] @ LyA, B: Ty[K, N] @ LyB, C: Ty[M, N] @ LyC):
        part_C = allo.matmul(A, B)
        C[:,:] = allo.allreduce(part_C, "+")
```

Fangtangtang:
There are still some unclear points in the current interface design.
For this specific GEMM example, the `allreduce` requires each `pk` to provide its `part_C`, but the final `C[:,:]` assignment should only happen in one of them (e.g., `pk == 0`).

In the current form, it looks like `C[:,:] = ...` would be executed in every `pk`.
A clearer way might be:

```python
LyA = Layout("S1S2")
LyB = Layout("S2S0")
LyC = Layout("S1S0")

@df.region()
def top():
    @df.kernel(mapping=[Pk, Pm, Pn])
    def gemm(A: Ty[M, K] @ LyA, B: Ty[K, N] @ LyB, C: Ty[M, N] @ LyC):
        part_C = allo.matmul(A, B)
        pk, pm, pn = df.get_pid()
        with allo.meta_if(pk == 0):
            C[:,:] = allo.allreduce(part_C, "+")
```

Here, the semantics of `allo.allreduce` implicitly gather all `part_C` values across different `pk`.

Fangtangtang:
For internal virtual graph construction, a reasonable approach is to introduce a dedicated **reduction node** for `allreduce`.

- All nodes that provide partial values will connect to this reduction node
- The reduction node connects to the node(s) that consume the final reduced value.

For example, in the GEMM case:

- Each `pk` produces its own `part_C` from `partial gemm`.
- All these `part_C` edges go into the `allreduce` reduction node.
- The reduction node then produces the final value, which is consumed only by the node `pk == 0` and output `C[:,:]`.

Thus, the virtual graph for the GEMM example may look like:

[...]

PR #432, opened 2025-10-04
[AIE][Frontend][IR][Builder] Add support for `memref` slice-based access and `gather`/`scatter` in frontend

Fangtangtang:
## Description ##
This PR partially fixes issue #425 and adds `gather` and `scatter` support to the frontend.

### Problems ###
Previously, slice-based access for `memref`/`tensor` types was not supported (see #425).

Besides, even with slice-based access, it was still difficult to express explicit gather/scatter patterns.
For example, the following code is not valid because `buffer[i, :, :]` cannot use a `meta_for` iterator as an index  (AIE backend requires `pid`/`meta_for` iterators to be used only for meta-programming and stream indexing):

```python
buffer: Ty[Pk, M, N]
with allo.meta_for(Pk) as i:
    buffer[i,:,:] = pipe[i].get()
```
### Proposed Solutions ###
For  slice-based access, this PR partially fixes that limitation by enabling proper slice-based access for `memref` objects. The core implementation introduces two new ops in the `allo` dialect: `load_slice` and `store_slice`. (More details in [this comment](https://github.com/cornell-zhang/allo/issues/425#issuecomment-3362324816))

Added frontend interfaces for `scatter` and `gather`:
```python
def gather(pipes: list)
```
```python
def scatter(buffer, pipes: list)
```
Here, `pipes` can be either:
- a **slice** of a pipe array (e.g., `df.gather(pipe[:])`), or
- an **explicit list** of pipe objects (e.g., `df.gather([pipe[0], pipe[1], pipe[2], pipe[3]])`).

### Examples ###
**1. Slice-based access**

```python
@df.region()
def top():
    @df.kernel(mapping=[1])
    def core(A: Ty[Pk, N], B: Ty[Pk, N]):
        for i in range(Pk):
            B[i, :] = A[i, :]
```

**2. `scatter` and `gather` usage**

```python
@df.region()
def top():
    pipe_1 = df.array(df.pipe(dtype=Ty, shape=(N,), depth=2), shape=(Pk,))
    pipe_2 = df.array(df.pipe(dtype=Ty, shape=(N,), depth=2), shape=(Pk,))

    @df.kernel(mapping=[1])
    def src(A: Ty[Pk, N]):
        df.scatter(A, pipe_1[:])

    @df.kernel(mapping=[Pk])
    def inter():
        pk = df.get_pid()
        pipe_2[pk].put(allo.add(pipe_1[pk].get(), 1))

    @df.kernel(mapping=[1])
    def dst(B: Ty[Pk, N]):
        B[:, :] = df.gather(pipe_2[:])
```

## Checklist ##

- [ ] Code is well-documented


Fangtangtang:
The failing test is `test_subview_systolic()` in `tests/test_systolic_array.py`.

The root cause is that some `kernel` parameters (e.g., `A_out`, `B_out`) are expected to be **passed by reference**, but in the parsed AST they are **loaded (passed by value)** instead.

reference code
```python
kernel(
    A_fifo[i, j], B_fifo[j, i], A_fifo[i, j + 1], B_fifo[j, i + 1], C, i, j
)
```

In the previous implementation, the IR builder directly created `subview`s for such arguments, treating them as references, which happened to produce correct behavior.

However, in the new implementation, I modified the conversion logic to avoid function signature type conflicts. The builder use value passing for such cases, which causes a correctness issue here.

From the AST-to-IR translation perspective, I think these arguments should indeed be treated as value-passed. Therefore, the issue may actually lie in how the AST is constructed, or possibly in how the source code of this specific test case is written?

Please let me know what the expected behavior is for such cases👀

Fangtangtang:
The issue that *during Python AST parsing, parameters passed by reference are gotten from a load node* requires a more systematic solution. 
Current fix for `subview` is not fundamental, for memref/tensor slice-based access, the current `LoadSliceOp` and `StoreSliceOp` still rely on node type handling and cannot fully avoid this problem.

A failing case:
```python
    norm = ExternalModule(
        top="single_row_layer_norm",
        impl_path="norm.cc",
        input_idx=[0, 1],
        output_idx=[2],
    )

    Ty = float32
    M, N = 4, 512

    @df.region()
    def top():
        @df.kernel(mapping=[1])
        def core(A: Ty[M, N], B: Ty[N], C: Ty[M, N]):
            for i in range(M):
                norm(A[i,:], B, C[i,:])
```

chhzh123:
> Current fix for subview is not fundamental, for memref/tensor slice-based access, the current LoadSliceOp and StoreSliceOp still rely on node type handling and cannot fully avoid this problem.

You mean the function interface is not consistent?

Fangtangtang:
I mean the AST parsing result is inconsistent with the function interface semantics, specifically on parameter passing.

Fangtangtang:
The current handling for `subview` (pass by reference by default and copy in/out only when type mismatch) feels more like a patch to the problem.😓

Fangtangtang:
I may have misunderstood the semantics of `ast.Load`. I’ll take another look into it.

chhzh123:
> The current handling for subview (pass by reference by default and copy in/out only when type mismatch) feels more like a patch to the problem.😓

I don't think you need to change the builder? Basically use subview and make sure the caller and callee interfaces are consistent. Then we can create explicit copies during the kernel injection pass.

Fangtangtang:
Yeah, exactly. I’ve fixed the `subview` issue, so if a parameter is passed through a `memref` access implemented with `subview`, it totally works fine now.

The problem is that `subview` can’t handle all kinds of `memref` accesses (that’s why I’m also introducing `load_slice` / `store_slice` in this PR)
One possible approach is to add similar special handling for all parameter passing cases, but I’m wondering if there’s a cleaner solution. Having fewer ad-hoc cases would be better for long-term maintenance and should adapt more naturally to future corner cases.

Fangtangtang:
> A failing case:
> 
> ```python
>     norm = ExternalModule(
>         top="single_row_layer_norm",
>         impl_path="norm.cc",
>         input_idx=[0, 1],
>         output_idx=[2],
>     )
> 
>     Ty = float32
>     M, N = 4, 512
> 
>     @df.region()
>     def top():
>         @df.kernel(mapping=[1])
>         def core(A: Ty[M, N], B: Ty[N], C: Ty[M, N]):
>             for i in range(M):
>                 norm(A[i,:], B, C[i,:])
> ```

In this case, the parameters for `norm` are gotten using `load_slice`, `subview` can not handle 'dynamic offset `i`'.
(https://github.com/cornell-zhang/allo/issues/425#issuecomment-3362324816)

chhzh123:
I see. Can we check whether the argument types are different in this case? If mismatched (like this kind of dynamic index), then we create explicit copies to make sure the types are consistent, just like type casting.

allo/ir/builder.py
```diff
@@ -2240,21 +2322,169 @@ class ExternalModule:
...
+                                offsets=offset_values,
+                                sizes=size_values,
+                                strides=stride_values,
+                            )
+                            allo_d.StreamPutOp(stream_op.result, [], op.result)
+                    if fn_name == "gather":
+                        return alloc_op
+                    return for_op
             # Allo library functions
             new_args = build_stmts(ctx, node.args)
```
Fangtangtang: There might be some difference here.
The main issue with dynamic indices is that during the builder’s visit to the argument list, it directly emits a `load_slice` (since the argument corresponds to a `Load` node in the AST).
For example, in the case above, the generated IR is:

```mlir
%0 = allo.load_slice %arg0[%arg3, %c0] sizes = [1, 512] strides = [512, 1] {from = "A"} : memref<4x512xf32> -> memref<512xf32>
%1 = allo.load_slice %arg2[%arg3, %c0] sizes = [1, 512] strides = [512, 1] {from = "C"} : memref<4x512xf32> -> memref<512xf32>
func.call @single_row_layer_norm(%0, %arg1, %1) : (memref<512xf32>, memref<512xf32>, memref<512xf32>) -> ()
```
However, the correct behavior should be that the inputs are loaded before the function call, and the outputs are stored back after the call.

zhangzhiru: This sounds like a bug.

Fangtangtang:
I actually had a perhaps similar idea, when the inferer analyzes the argument list for a call op, we could annotate the AST node (the Load node) with an additional attribute to distinguish whether the parameter is an input or an output.🤔

allo/ir/builder.py
Fangtangtang: Exactly, it is.

chhzh123: This `load_slice` is introduced by this PR, right?

chhzh123:
It seems to be some additional transformations. Do we need to implement this in the builder?

allo/ir/builder.py
Fangtangtang: yes

allo/ir/builder.py
```diff
@@ -2240,21 +2322,169 @@ class ExternalModule:
...
             if isinstance(obj, (IPModule, ExternalModule)):
+                input_idx = obj.input_idx if isinstance(obj, ExternalModule) else None
```
Fangtangtang, line 2429: The annotation can be added in infer. 
The builder may need to perform corresponding transformations when constructing the argument list for the Call operation (similar to the logic around lines 2462–2506 in the builder.py).

Fangtangtang:
> I may have misunderstood the semantics of `ast.Load`. I’ll take another look into it.

My fault, I confused python’s parameter passing semantics. The `expr_context = Load` in function call arguments means “fetching the object”, but it doesn’t imply that the object itself is immutable.
So my transformation using `load_slice` / `store_slice` for handling passed parameters is semantically incorrect.

Fangtangtang:
Can I just add a guard in the builder or type inferer to invalidate this usage (i.e., disallow dynamic slices as call operation arguments)? 😶‍🌫️

From the AST-to-IR semantic correctness perspective, it should ideally be lowered to a `subview`. The reason we can’t use `memref.subview` for dynamic slices right now is that dynamic offset/size/stride support in `memref.subview` is still not fully compatible with many MLIR builtin passes (this might improve in the coming years).
Introducing `load_slice` and `store_slice` doesn’t have correctness issues when used under regular load/store contexts, they can simply serve as a workaround for handling some dynamic slices cases.

tests/dataflow/aie/test_norm.py
```diff
@@ -114,7 +113,42 @@ def core(A: Ty[M, N] @ LyA, B: Ty[N] @ Ly, C: Ty[M, N] @ LyA):
...
+    Ty = float32
+    M, N = 4, 512
+
+    @df.region()
+    def top():
+        @df.kernel(mapping=[1])
+        def core(A: Ty[M, N], B: Ty[N], C: Ty[M, N]):
+            for i in range(M):
+                # [NOTE]: test using buffer slice as customized external kernel arguments
+                norm(A[i], B, C[i])
```
Fangtangtang, line 134: The layer norm failing case above actually has a valid equivalent expression (see lines 128–134 here).

chhzh123:
> Can I just add a guard in the builder or type inferer to invalidate this usage

Sure, we can add a guard if this feature is not necessary for implementing your kernels

allo/ir/builder.py
```diff
@@ -1163,26 +1163,105 @@ def build_slices(ctx: ASTContext, node: ast.Subscript, in_shape: ShapedType):
...
+                    continue
+                raise ValueError(
+                    f"Index type ({type(index)}) not supported. Dynamic offset is incompatible with many builtin opt passes."
+                )
+            else:
+                raise ValueError(f"Unsupported slice index type ({type(index)})")
+        return offsets, sizes, strides, static_offsets, static_sizes, static_strides
+
+    @staticmethod
+    def build_symbolic_slices(
```
chhzh123: What is a symbolic slice? When we will use it

allo/ir/builder.py
```diff
@@ -1274,29 +1360,82 @@ def build_memory_access(
...
+                    )
+                else:
+                    op = allo_d.StoreSliceOp(
+                        val.result,
+                        value.result,
+                        offsets=offset_values,
+                        sizes=static_sizes,
+                        strides=stride_values,
+                        ip=ctx.get_ip(),
+                    )
```
chhzh123: Do we still need this new operation?

chhzh123: Can we use some combination of upstream operations instead?

mlir/include/allo/Conversion/Passes.h
```diff
@@ -21,13 +21,18 @@ std::unique_ptr<OperationPass<ModuleOp>> createFixedPointToIntegerPass();
 std::unique_ptr<OperationPass<ModuleOp>> createLowerCompositeTypePass();
 std::unique_ptr<OperationPass<ModuleOp>> createLowerBitOpsPass();
 std::unique_ptr<OperationPass<ModuleOp>> createLowerTransformLayoutOpsPass();
+std::unique_ptr<OperationPass<ModuleOp>> createLowerStoreSliceOpsPass();
+std::unique_ptr<OperationPass<ModuleOp>> createLowerLoadSliceOpsPass();
+std::unique_ptr<OperationPass<ModuleOp>> createLowerMemrefOpsPass();
```
chhzh123: Same, can we reuse the upstream operations and passes?

allo/ir/builder.py
Fangtangtang: yes, we need these for current `gather`/`scatter` frontend interface lowering and the examples in `test_slice.py`

allo/ir/builder.py
Fangtangtang: For `df.gather(pipe[:])` it is used to build the loop bound (to load the pipe[i] slices to a buffer) for `pipe[:]`.
used in line 2359.

Fangtangtang: The name is confusing, I can simply inline it and add some comments

allo/ir/builder.py
Fangtangtang: > Can we use some combination of upstream operations instead?

From builtin dialects? so I'll directly lower the instructions to nested loops of load/stores in the IR? (copy to/from subview won't work as discussed above)

chhzh123: ok, but then you need to analyze which arguments are read/written during the build process?

Fangtangtang: That’s actually two separate things.
During memory access, I decide based on the AST node’s `expr_context` whether it’s a load or store, and then build either a `load_slice` or `store_slice`. This works correctly for normal load/store operations.
e.g., for the examples in `test_slice.py`
```python
        @df.kernel(mapping=[1])
        def core(A: Ty[N], B: Ty[Pk, N]):
            for i in range(0, Pk, 2):
                B[i, :] = allo.add(A, 1)
            for i in range(1, Pk, 2):
                B[i, :] = allo.add(A, -1)
```
However, since function call arguments have `expr_context=load` but are actually passed as objects rather than loaded values, they shouldn’t be built as `load_slice`. So I invalidated this usage for now.

Fangtangtang: `load_slice` and `store_slice` are not dedicated for function call arguments, it's for usage like `B[i, :]`.
When `B[i, :]` is used in the argument list, it leads to correctness issue, so I'll raise error in the builder for such cases now.

allo/ir/builder.py
```diff
@@ -1274,29 +1360,82 @@ def build_memory_access(
...
+            result_sizes = []
+            result_strides = []
+            for idx_, size in enumerate(static_sizes):
+                if size > 1:
+                    result_sizes.append(size)
+                    result_strides.append(new_strides[idx_])
+            if len(offsets) > 0:
+                assert (
+                    ctx.in_call_arg_list == 0
+                ), "Dynamic slices as call operation arguments are not supported yet"
```
Fangtangtang: Here is the guard

Fangtangtang:
Emm, I also agree that we should try to keep the Allo dialect as clean and minimal as possible, avoiding adding new operations unless really necessary.
The `load/store_slice` ops were mainly introduced to compensate for some limitations in the `memref` dialect. For now, the functionality they directly add doesn’t seem that critical. (Please check the examples in `test_slice.py` and see if you think they’re really necessary to support.)

---
The `load/store` dynamic indexed slice pattern was used to express scatter/gather behavior, which in turn is needed to support the allreduce pattern. And all of this ultimately serves to support the GEMM interface. #406
I think adding `allreduce`, `scatter`, and `gather` operations in the Allo dialect is necessary to properly support the GEMM interface.
If we want to avoid directly introducing `load/store_slice` into the dialect, I think it’s also reasonable to hide them inside the `scatter/gather` operations instead. (but examples in `test_slice.py` cannot be supported in this case)

Fangtangtang:
> The reason we can’t use `memref.subview` for dynamic slices right now is that dynamic offset/size/stride support in `memref.subview` is still not fully compatible with many MLIR builtin passes (this might improve in the coming years). Introducing `load_slice` and `store_slice` doesn’t have correctness issues when used under regular load/store contexts, they can simply serve as a workaround for handling some dynamic slices cases.

Here’s an example of using a dynamic offset in `memref.subview`.

```python
def _test_store_slice():

    Ty = int16
    N = 32
    Pk = 4

    @df.region()
    def top_v1():

        @df.kernel(mapping=[1])
        def core(A: Ty[N], B: Ty[Pk, N]):
            for i in range(Pk):
                B[i, :] = A

    A = np.random.randint(0, 64, (N)).astype(np.int16)
    B = np.zeros((Pk, N)).astype(np.int16)

    mod = df.build(top_v1, target="aie")
```

Error message:
```txt
Error: failed to run MLIR lower pipeline, printing module...
#map = affine_map<() -> (0)>
#map1 = affine_map<() -> (4)>
"builtin.module"() ({
  "func.func"() <{function_type = (memref<32xi16>, memref<4x32xi16>) -> (), sym_name = "core_0"}> ({
  ^bb0(%arg0: memref<32xi16>, %arg1: memref<4x32xi16>):
    "affine.for"() <{lowerBoundMap = #map, operandSegmentSizes = array<i32: 0, 0, 0>, step = 1 : index, upperBoundMap = #map1}> ({
    ^bb0(%arg2: index):
      %0 = "memref.subview"(%arg1, %arg2) <{operandSegmentSizes = array<i32: 1, 1, 0, 0>, static_offsets = array<i64: -1, 0>, static_sizes = array<i64: 1, 32>, static_strides = array<i64: 1, 1>}> : (memref<4x32xi16>, index) -> memref<32xi16, strided<[1], offset: -32>>
      "memref.copy"(%arg0, %0) {to = "B"} : (memref<32xi16>, memref<32xi16, strided<[1], offset: -32>>) -> ()
      "affine.yield"() : () -> ()
    }) {loop_name = "i", op_name = "S_i_0"} : () -> ()
    "func.return"() : () -> ()
  }) {df.kernel, itypes = "ss", otypes = "", stypes = "__", tag = "core_()"} : () -> ()
  "func.func"() <{function_type = () -> (), sym_name = "top_v1"}> ({
    "func.return"() : () -> ()
  }) {itypes = "", otypes = ""} : () -> ()
}) : () -> ()
```
Error occurred in `dataflow.py`
```txt
line 265, in _build_top
    mlir_pass_manager.parse(pipeline).run(s.module.operation)
allo._mlir._mlir_libs._site_initialize.<locals>.MLIRError: Failure while executing pass pipeline:
error: unknown: expected 0 dynamic offset values
 note: unknown: see current operation: %0 = "memref.subview"(%arg1, %arg2) <{operandSegmentSizes = array<i32: 1, 1, 0, 0>, static_offsets = array<i64: -1, 0>, static_sizes = array<i64: 1, 32>, static_strides = array<i64: 1, 1>}> : (memref<4x32xi16>, index) -> memref<32xi16, strided<[1], offset: -32>>
```

Fangtangtang:
I’ve decided to refactor. I’ll add `scatter` and `gather` operations in the Allo dialect and avoid directly introducing `load/store_slice` into it. 
I’ll close this PR for now.

Fangtangtang:
Found a valid syntax for `memref.subview` with dynamic offset. no need for `load_slice` / `store_slice` and can cover all `test_slice.py` examples, so I’m reopening the PR.
I think this is great news. 🤩

zhangzhiru:
Great news indeed. What's the correct syntax?

Fangtangtang:
For example, in the case that caused the error above, it **should not** be

```
%0 = "memref.subview"(%arg1, %arg2) <{operandSegmentSizes = array<i32: 1, 1, 0, 0>, static_offsets = array<i64: -1, 0>, static_sizes = array<i64: 1, 32>, static_strides = array<i64: 1, 1>}> : (memref<4x32xi16>, index) -> memref<32xi16, strided<[1], offset: -32>>
```
but
```
%0 = "memref.subview"(%arg1, %arg2) <{operandSegmentSizes = array<i32: 1, 1, 0, 0>, static_offsets = array<i64: -9223372036854775808, 0>, static_sizes = array<i64: 1, 32>, static_strides = array<i64: 1, 1>}> : (memref<4x32xi16>, index) -> memref<32xi16, strided<[1], offset: ?>>
```

using `-9223372036854775808` to indicate *dynamic* and setting the `offset` in the `memref` type to `?`.

chhzh123:
I thought the issue was that many MLIR passes do not support dynamic indices? There are no other problems if we make this dynamic?

Fangtangtang:
> I thought the issue was that many MLIR passes do not support dynamic indices?

Not exactly. The error I mentioned earlier was actually caused by the syntax issue, but I misunderstood the error message. Since the operation built successfully but the pass failed, I initially thought it was because the pass didn’t support dynamic indices.

> There are no other problems if we make this dynamic?

One potential issue is the `?` in the memref type. It can lead to type mismatches. As mentioned earlier, if the result of a `subview` has a `?` in its `memref` type but the function interface declares a static `memref` type, the MLIR verifier will reject it because the two memref types don’t match.

chhzh123:
I see. That's my initial thought. Then we need to update the function interface, but it's a bit tricky, since you also need to update the types of the inner variables (similar to what [`.partition()`](https://github.com/cornell-zhang/allo/blob/main/allo/customize.py#L272-L657) primitive does)

Fangtangtang:
> As mentioned earlier, if the result of a `subview` has a `?` in its `memref` type but the function interface declares a static `memref` type, the MLIR verifier will reject it because the two memref types don’t match.

This issue was fixed in this PR. The solution is to allocate a buffer with static `memref` type for passing arguments. for memrefs that are read inside the function, we copy them to the buffer before the call, and for those that are written inside the function, we copy them back after the call. 

The correctness should be fine, but this may introduce some overhead. I’m not sure whether later passes will automatically optimize it away.

Fangtangtang:
Just to add, for AIE backend, we chose not to update the interface types to dynamic mainly because of an issue with the MLIR-AIE. As far as I remember, it doesn’t support dynamic `memref` types in function interfaces.

---

Recording the reproducible error here. [top.zip](https://github.com/user-attachments/files/22867853/top.zip)

run `aiecc.py --alloc-scheme=basic-sequential --aie-generate-xclbin --no-compile-host --xclbin-name=build/final.xclbin --no-xchesscc --no-xbridge --peano $PEANO_INSTALL_DIR --aie-generate-npu-insts --npu-insts-name=insts.txt top.mlir`

error message:
```txt
 MLIR compilation: ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━   0% -:--:-- 0:00:00 0/1 1 Worker/top.prj/top.mlir.prj/input_with_addresses.mlir:388:7: error: 'llvm.call' op 'add_i16_vector' does not reference a valid LLVM function
      func.call @add_i16_vector(%fifo_4_buff_0, %subview, %_anonymous3) {lib = "add_i16_vector"} : (memref<16x16xi16>, memref<16x16xi16, strided<[16, 1], offset: ?>>, memref<16x16xi16>) -> ()
      ^
/top.prj/top.mlir.prj/input_with_addresses.mlir:388:7: note: see current operation: "llvm.call"(%795, %796, %797) <{CConv = #llvm.cconv<ccc>, TailCallKind = #llvm.tailcallkind<none>, callee = @add_i16_vector, fastmathFlags = #llvm.fastmath<none>, op_bundle_sizes = array<i32>, operandSegmentSizes = array<i32: 3, 0>}> {lib = "add_i16_vector"} : (!llvm.ptr, !llvm.ptr, !llvm.ptr) -> ()
Error encountered while running: aie-opt --pass-pipeline=builtin.module(aie.device(aie-localize-locks,aie-normalize-address-spaces),aie-standard-lowering,aiex-standard-lowering,canonicalize,cse,convert-vector-to-llvm,expand-strided-metadata,lower-affine,convert-math-to-llvm,convert-index-to-llvm,arith-expand,convert-arith-to-llvm,finalize-memref-to-llvm,convert-func-to-llvm{ use-bare-ptr-memref-call-conv=1 },convert-cf-to-llvm,canonicalize,cse) /top.prj/top.mlir.prj/input_with_addresses.mlir -o /top.prj/top.mlir.prj/input_opt_with_addresses.mlir
```

chhzh123:
> Just to add, for AIE backend, we chose not to update the interface types to dynamic mainly because of an issue with the MLIR-AIE.

Sure, this makes sense

zhangzhiru:
> Just to add, for AIE backend, we chose not to update the interface types to dynamic mainly because of an issue with the MLIR-AIE. As far as I remember, it doesn’t support dynamic memref types in function interfaces.

Shall we open separate issue(s) for these known problems so we don’t lose track of them? We can revisit once the dependent issues are fixed.

chhzh123:
@Fangtangtang Let me know when you've finished the changes

Fangtangtang:
finished

Issue #441, opened 2025-10-23
[Feature][BUG] Clarify Grammar Definition; Fix Broken Testcases after PR #427


Fangtangtang:
PR #427 addressed several issues related to variable scoping but did not fully update all test cases.
After that PR, some existing test cases are now reported as syntax errors and need to be fixed.

Some of the affected test cases include:

- `tests/dataflow/test_multi_cache_gemm.py`
- `tests/dataflow/test_unified_systolic.py`
- `tests/dataflow/test_daisy_chain_gemm.py`
- `tests/dataflow/test_unified_systolic.py`

In addition, we have never clearly defined the grammar rules for variable *declaration*, *assignment*, and *scope*. The current logic for type inference and IR building is quite messy. We need to clarify Allo’s grammar definition and clean up the related logic.

**Describe the solution you'd like**

We should:

- [ ] Clarify Allo’s grammar definition for variable declaration, assignment, and scoping.
- [ ] Clean up the logic for type inference and IR building.
- [ ] Fix all buggy test cases according to the clarified grammar definition.
- [ ] *Add dataflow test cases to the CI pipeline.

Fangtangtang:
## Proposed Grammar Definition for Variable Declaration, Assignment, and Scoping

---
Some context for different assign statements:
```python
a = 1 		# ast.Assign
a: int32 = 1	# ast.AnnAssign: assign with type annotation
a += 1		# ast.AugAssign
```
---
**Declaration**
- A variable is *declared* the first time it appears as the left-hand side of an `Assign` or `AnnAssign` statement within a function (kernel) body. 
- Variable re-declaration is not allowed.

**Assignment**
- Any subsequent appearance of a variable on the left-hand side of `Assign`, `AnnAssign`, or `AugAssign` after its first declaration is treated as an assignment.
- A variable must be declared in the current or enclosing scopes before it can be assigned.
- Reassignment cannot change the declared variable’s type or shape. The value and the assigned variable must be type compatible.

**Scoping**
- Scoping follows C++-style.
- Each `if` branch or `for` loop (including `meta_if` and `meta_for`) introduces a new subscope.
- Variables declared in a scope are visible within that scope and all its nested subscopes.
- Forward reference (using a variable before its declaration) is not allowed.

Issue #445, opened 2025-10-23
[BUG] Unsupported Name 'packed_tmp' error


silvenachen:
**Describe the bug**
Allo throws `RuntimeError: Unsupported Name 'packed_tmp'` during IR building when `packed_tmp` is used.

**To Reproduce**
multiple test cases under the dataflow path, including `tests/dataflow/test_multi_cache_gemm.py` and `tests/dataflow/test_daisy_chain_gemm.py`
Simply run these testcases with `python3 test_multi_cache_gemm.py` will reproduce the issues.

**Buggy output**
```
/export/hdd/scratch/sli3079/allo/allo/_mlir/exceptions.py:76: RuntimeWarning: 
[Data Type] NumPy does not support bitwidths larger than 512, so you cannot use NumPy 
for simulation, but you can still generate the corresponding MLIR module.
  warnings.warn(self.message, category=self.category)

Traceback (most recent call last):
  File "/export/hdd/scratch/sli3079/allo/allo/ir/infer.py", line 1330, in visit_stmts
    results.append(visit_stmt(ctx, stmt))
  File "/export/hdd/scratch/sli3079/allo/allo/ir/visitor.py", line 227, in __call__
    res = method(ctx, node)
  File "/export/hdd/scratch/sli3079/allo/allo/ir/infer.py", line 318, in visit_Assign
    rhs = visit_stmt(ctx, node.value)
  File "/export/hdd/scratch/sli3079/allo/allo/ir/visitor.py", line 227, in __call__
    res = method(ctx, node)
  File "/export/hdd/scratch/sli3079/allo/allo/ir/infer.py", line 442, in visit_Subscript
    value = visit_stmt(ctx, node.value)
  File "/export/hdd/scratch/sli3079/allo/allo/ir/visitor.py", line 227, in __call__
    res = method(ctx, node)
  File "/export/hdd/scratch/sli3079/allo/allo/ir/infer.py", line 135, in visit_Name
    raise RuntimeError(f"Unsupported Name `{node.id}`")
RuntimeError: Unsupported Name `packed_tmp`

Error: Unsupported Name `packed_tmp`
╭────────────────── Traceback (most recent call last): Line: 122 ───────────────────╮
│  117:             packed_c: UInt(M * 8) = 0                                       │
│  118:             for m in range(Rt):                                             │
│  119:                 if m == i - 1:                                              │
│  120:                     packed_c[m * 8:(m + 1) * 8] = c                         │
[...]

Issue #444, opened 2025-10-23
[BUG] [Dataflow] Single stream shared by multiple producers/consumers in dataflow desing cooperative gemv


silvenachen:
**Describe the bug**
For the design under /tests/dataflow/test_cooperative_gemv, Allo generates a single stream shared by multiple kernel instances. This violates HLS dataflow constraints that require one producer and one consumer per stream. The tests did not pass presynthesis, even when I manually added the missing array partitioning for buf0, buf1, and buf2. 

**To Reproduce**
```
# Copyright Allo authors. All Rights Reserved.
# SPDX-License-Identifier: Apache-2.0

from allo.ir.types import float32
import allo.dataflow as df
import allo.backend.hls as hls
import numpy as np

Ty = float32
M, N = 16, 16
P0 = 2
Mt = M // P0


@df.region()
def top():
    pipe = df.pipe(dtype=Ty, shape=(Mt,), depth=2)

    @df.kernel(mapping=[P0])
    def gemv0(A: Ty[M, N], x: Ty[N]):
        pi = df.get_pid()
        y_out: Ty[Mt] = 0
        for m in range(pi * Mt, (pi + 1) * Mt):
            y_acc: Ty = 0
            for n in range(N // 2):
                y_acc += A[m, n] * x[n]
            y_out[m - pi * Mt] = y_acc
        pipe.put(y_out)

    @df.kernel(mapping=[P0])
    def gemv1(A: Ty[M, N], x: Ty[N], y: Ty[M]):
        pi = df.get_pid()
        y_out: Ty[Mt] = 0
        for m in range(pi * Mt, (pi + 1) * Mt):
            y_acc: Ty = 0
            for n in range(N // 2, N):
                y_acc += A[m, n] * x[n]
            y_out[m - pi * Mt] = y_acc
        y_prev: Ty[Mt] = pipe.get()
        for m in range(pi * Mt, (pi + 1) * Mt):
            y[m] = y_out[m - pi * Mt] + y_prev[m - pi * Mt]


def test_cooperative_gemv():
    A = np.random.rand(M, N).astype(np.float32)
    B = np.random.rand(
        N,
    ).astype(np.float32)
    C = np.zeros((M,), dtype=np.float32)
    mod = df.build(top)
    if hls.is_available("vitis_hls"):
        # mod(A, B, C)
        # np.testing.assert_allclose(C, np.dot(A, B), atol=1e-5)
        # print("Passed!")
[...]

PR #447, opened 2025-10-25
[IR][Builder] Refactor variable and scoping logic

Fangtangtang:
## Description ##
This PR aims to fix issues #441 and #445.
It mainly addresses problems related to *Variable Declaration, Assignment, and Scoping*, updating the type inference and IR builder according to the [new syntax definition](https://github.com/cornell-zhang/allo/issues/441#issuecomment-3434654299).

Additionally, it fixes test cases that caused syntax errors under the new rules and adds several basic builder test cases.

> [NOTE]: This PR is currently opened for testing and further discussion.  **NOT READY TO MERGE**.

### Problems ###
Discussed in #441 and #445.

### Proposed Solutions ###
Discussed in #441

## Checklist ##

- [ ] Code is well-documented

### DIFF
--- a/allo/passes.py
+++ b/allo/passes.py
@@ -698,11 +698,13 @@ def analyze_use_def(mod):
                 if "name" in op.attributes:
                     buf_name = f"{func_name}:{op.attributes['name'].value}"
                 elif " = " in str(op):
+                    if "from" in op.attributes:
+                        continue
                     op_name = str(op).split(" = ", maxsplit=1)[0]
                     buf_name = f"{func_name}:{op_name}"
                     op.attributes["name"] = StringAttr.get(op_name, context=mod.context)
                 else:
-                    # call op does not have return value
+                    # op without return value
                     continue
                 uf_add(buf_name)
                 add_use(op.result, buf_name)


Fangtangtang:
There are a few parts that I think we still need to supplement (either in this PR or in follow-up ones):

- Add/update syntax docs
- Fix broken test cases and buggy passes. Some of the existing test cases are now invalid (not due to syntax errors). I’ve added comments to some of the ones I’ve found so far.
- Integrate all test cases in the repo into CI.
  	- Some test cases (especially under `tests/dataflow/`) have been broken since earlier commits, and I currently can’t pinpoint all the issues, these will need to be fixed gradually. Mark some of them here as a reminder
		- tests/dataflow/test_cooperative_gemv.py #444
		- tests/dataflow/test_smith_waterman_systolic.py
		- tests/dataflow/test_unified_systolic.py
		- tests/dataflow/test_systolic_conv.py

allo/ir/builder.py
```diff
@@ -108,32 +108,46 @@ def __call__(self, ctx: ASTContext, node: ast.AST, file_name: str = None, **kwar
 
 # pylint: disable=too-many-public-methods
 class ASTTransformer(ASTBuilder):
+    # to resolve global variable naming conflict
+    global_var_cnt: dict[str, int] = {}
```
chhzh123, line 112: Why do we need to count the number of each global var? Can't we just record whether a var is defined or not?

allo/ir/builder.py
```diff
@@ -837,170 +863,193 @@ def build_indices(ctx: ASTContext, node: ast.AST, enable_affine: bool = True):
...
-                # [NOTE]: wired bug, `ip=ctx.get_ip()` not work
-                with ctx.get_ip():
-                    fill_op = linalg_d.fill(val.result, outs=[rhs.result])
-                    if ctx.enable_tensor:
-                        rhs = fill_op.owner
-            else:
-                constant = MockConstant(node.value.value, ctx)
-                cast = ASTTransformer.build_cast_op(
-                    ctx, constant, node.value.dtype, node.dtype, node.value.shape
+        conversion_enabled = False
```
chhzh123, line 892: What kind of conversion? You meant type casting?

allo/ir/builder.py
```diff
@@ -837,170 +863,193 @@ def build_indices(ctx: ASTContext, node: ast.AST, enable_affine: bool = True):
...
+                        else:
+                            alloc_op = rhs
+                        ctx.buffers[target.id] = alloc_op
+                        if isinstance(alloc_op, OpView):
+                            alloc_op.attributes["name"] = StringAttr.get(target.id)
+                        ctx.put_symbol(name=target.id, val=alloc_op)
+                else:
+                    assert idx is None, "Not Supported"
+                    ctx.buffers[target.id] = rhs
+                    # FIXME (Shihan): GetGlobalOp has a "name" attribute, which may
```
chhzh123: may...?

allo/ir/builder.py
```diff
@@ -1255,7 +1305,8 @@ def build_tensor_access(
                     strides=[],
                     ip=ctx.get_ip(),
                 )
-        if isinstance(node.slice, (ast.Index, ast.Tuple)):
+        # access a element
```
chhzh123: an element

allo/ir/builder.py
```diff
@@ -2582,6 +2597,7 @@ class ExternalModule:
                 arg_results = [arg.result for arg in new_args]
                 input_types = [arg.type for arg in arg_results]
                 output_types = [input_types[0]]
+                print(f"{fn_name}_{abs(hash(node))}")
```
chhzh123: delete

allo/ir/infer.py
```diff
@@ -253,7 +257,7 @@ def visit_For(ctx: ASTContext, node: ast.For):
 
     @staticmethod
     def visit_broadcast(
-        ctx: ASTContext, lhs: ast.expr, rhs: ast.expr, match_lhs: bool = False
+        ctx: ASTContext, lhs: ast.expr, rhs_shape: list[int], match_lhs: bool = False
```
chhzh123: It's a bit weird here. Why do we want to pass in a list of int instead?

allo/ir/infer.py
```diff
@@ -715,6 +775,7 @@ def get_predicate_list(predicate_raw, pid_map):
                 )
                 # update shape
                 arg.shape = arg.dtensor.get_local_shape()
+                assert ctx.get_symbol(name=arg.arg, allow_missing=True) is None
```
chhzh123: Add necessary error message

allo/ir/infer.py
```diff
@@ -742,6 +803,7 @@ def get_predicate_list(predicate_raw, pid_map):
                     node.returns.dtype, node.returns.shape, node.returns.spec = (
                         TypeInferer.visit_type_hint(ctx, node.returns)
                     )
+                assert ctx.get_symbol(name=func_name, allow_missing=True) is None
```
chhzh123: Similarly, add error messages

tests/autoscheduler/test_passes.py
```diff
@@ -20,11 +20,11 @@
 @pytest.mark.parametrize("kind", kinds)
 def test_simple(debug_point, kind):
     def simple(v: int32[10, 10]) -> int32[10, 10]:
-        def stageA(v: int32[10, 10]) -> int32[10, 10]:
+        def stageA(v_: int32[10, 10]) -> int32[10, 10]:
             A: int32[10, 10]
             for j in range(10):
                 for i in range(10):
-                    A[i, j] = i + j + v[i, j]
+                    A[i, j] = i + j + v_[i, j]
```
chhzh123, line 27: What's the issue here?

tests/dataflow/aie/test_cannon.py
```diff
@@ -46,11 +46,13 @@ def cannon(C: Ty[M, N] @ LyC):
...
-                C[:, :] += allo.matmul(A_out, B_out)
-                A_pipe[(pi - 1) % P, pj].put(A_out)
-                B_pipe[pi, (pj - 1) % P].put(B_out)
+                # Using an already created variable (A_out) is also valid,
+                # but the compiler currently performs better optimization when using a new variable here.
+                A_out_: Ty[m, k] = A_pipe[pi, pj].get()
+                B_out_: Ty[k, n] = B_pipe[pi, pj].get()
+                C[:, :] += allo.matmul(A_out_, B_out_)
+                A_pipe[(pi - 1) % P, pj].put(A_out_)
+                B_pipe[pi, (pj - 1) % P].put(B_out_)
```
chhzh123, line 61: Why defining a new variable performs better? In terms of what?

tests/test_builder.py
```diff
@@ -859,31 +959,210 @@ def kernel4(n: int32) -> int32:
...
         s: int32 = 0
-        for i in range(n):  # outer loop, iterator is i
-            for i in range(n):  # inner loop reuses the same name i -> shadowing outer i
-                s = s + i  # here i refers to the inner loop variable
-            s = s + i  # here i refers to the outer loop variable
+        for i in range(n):
+            for j in range(n):
+                s = s + j
+            s = s + i
         return s
```
chhzh123, line 969: so there's no longer shadowing?

tests/test_schedule_compose.py
```diff
@@ -25,7 +25,7 @@ def kernel(A: int32) -> int32:
         return E
 
     s = allo.customize(kernel)
-    assert s.get_equivalent_variables("kernel:D") == set(
+    assert s.get_equivalent_variables("kernel:D") >= set(
```
chhzh123: Why `>=`? The dataflow graph changed?

allo/ir/builder.py
Fangtangtang: used for np.array. e.g., rhs of `tmp: int32[2, 2] = [[1, 2], [3, 4]]`. The rhs is will be a global var and the name is the same as the lhs var `memref.global "private" @tmp : memref<2x2xi32> = dense<[[1, 2], [3, 4]]>`.

If we have 
```
    def kernel3() -> int32:
        # declaration: type annotation required
        tmp: int32[2, 2] = [[1, 2], [3, 4]]
        ...
        # assignment
        tmp = [[2, 1], [3, 4]]
        return tmp[0, 0]
```
This two global var will have name conflict. So I count the number to avoid the conflict.

--- 

**why not global counter for all vars?**
I want to keep the name of global var used to initiate a variable when it is defined to be the same as the lhs varible name. 

This is actually a patch related to `partition` primitive. The the patitioned variable is initiated with a constant. The partition will 'propagate' to the rhs value and the global variable. (This also have something to do with another issue with `GetGlobalOp` in builder L964)
Seems that some internal pass rely on string matching of the "name" attribute to locate variables that need to be partitioned. (So this also means that if another global variable is used to assign a variable that needs to be partitioned, the pass will fail.)

allo/ir/builder.py
Fangtangtang: "may have assignment conflict" 😅

Part of the issue is mentioned [above](https://github.com/cornell-zhang/allo/pull/447#discussion_r2463009110). 
If different kernel functions use np.array to initiate variable with same name when it is defined, this also have a global var naming conflict that we need to use renaming to avoid. Then, if such a variable needs to be partitioned, a similar issue arises.
So I simply invalidate such cases in the builder here.

An 'invalidated' example:
```python3
    def kernel4():
        def a():
            # declaration: type annotation required
            tmp: int32[2] = [1, 2]

        def b() -> int32:
            tmp: int32[2] = [3, 4]
            return tmp[0]
```

allo/ir/builder.py
Fangtangtang: type casting, broadcasting

tests/autoscheduler/test_passes.py
Fangtangtang: the argument name `v` in `stageA` conflict with `v` in `simple`

tests/dataflow/aie/test_cannon.py
Fangtangtang: My 'buffer elimination pass' is not that strong. Defining a new variable simplifies the use/def chain, so the pass can better identify redundant ones

tests/test_schedule_compose.py
Fangtangtang: intermidiate results are also included

tests/test_builder.py
Fangtangtang: Ohhh, that’s a good point!!

Besides the lhs of the first assignment statement, loop iterators introduce new variables, and each new variable gets a **symbol** in the current scope. Function arguments and function names also introduce **symbols** into the scope.
Currently, my approach is to disallow duplicate symbol names in the context. But we can certainly change it to support shadowing instead.

Totally dependes on how we want to define it.

allo/ir/infer.py
Fangtangtang: Because I wanted to reuse the function, but not all `rhs`s are ast nodes. In fact, the function only uses `lhs.shape` and `rhs.shape`. If we want to keep the parameters consistent, we could also change `lhs` to `lhs_shape` (which is also a list of ints).

Fangtangtang:
**[update]**: Fix fragile string-based MLIR type construction in `builder`

Replaced `MLIRType.parse(...)` with direct `MemRefType.get(...)` to avoid crashes caused by string parsing. (Addresses [the issue](https://github.com/Fangtangtang/allo/issues/11) where MLIR type parsing could fail under different environments.)

tests/test_builder.py
chhzh123: so I wonder in your new implementation, what will happen when defining `i` again in the inner loop?

tests/dataflow/aie/test_cannon.py
chhzh123: ok, maybe make more explanations in the comment or leave a TODO item

allo/ir/infer.py
chhzh123: yeah, making them consistent will be better

allo/ir/builder.py
chhzh123: type casting can be turned off, but broadcasting should always be enabled?

tests/test_builder.py
Fangtangtang: now it raise an error `Please choose a different name for the loop iterator.`

allo/ir/builder.py
chhzh123: I still don't get why counting helps your case. In your example, `global_var_cnt["tmp"] = 2`?

Fangtangtang: yes

Fangtangtang: It counts the number of global variables like `memref.global "private" @tmp : memref<2x2xi32> = dense<[[1, 2], [3, 4]]>.` assigned to local varibale named `tmp`

allo/ir/builder.py
Fangtangtang: Actually, `conversion_enabled`is a guard. Currently, it is set to `false` in two cases:

1. When the RHS is a constant tensor that has already been converted according to the LHS type. **casting/broadcasting not needed**
2. When the RHS is an tuple that needs to be unpacked. **casting/broadcasting is not yet supported**

chhzh123: ok, so this is not to turn off all the casting that may potentially benefit naive AIE kernel codegen? since sometimes addition like `i16+i16` casting to `i17` is unnecessary.

tests/test_schedule_compose.py
chhzh123: Will this impact the `.partition()` algorithm?

allo/ir/builder.py
Fangtangtang: yeah, this doesn’t perform any optimization. It just collect the previously scattered assignment logic together.

tests/test_schedule_compose.py
Fangtangtang: should these intermidiate results be partitioned as well? 👀

There are some test cases for `.partition()`, seem that it works fine. I'm not sure whether that's just because the test cases do not cover corner cases

chhzh123: Can you print out what the `s.get_equivalent_variables` are in this case? What're the variables other than `["foo2:0", "foo:0", "kernel:D"]`

Fangtangtang: `{'kernel:D', 'foo2:0', 'kernel:%21', 'foo:0'}`

chhzh123: What is this `kernel:%21`? Can you also attach the whole MLIR code?

Fangtangtang:
```mlir
module {
  func.func @foo2(%arg0: i32) -> i32 attributes {itypes = "s", otypes = "s"} {
    %0 = arith.extsi %arg0 : i32 to i33
    %c1_i32 = arith.constant 1 : i32
    %1 = arith.extsi %c1_i32 : i32 to i33
    %2 = arith.addi %0, %1 : i33
    %3 = arith.trunci %2 : i33 to i32
    %alloc = memref.alloc() {name = "B"} : memref<i32>
    affine.store %3, %alloc[] {to = "B"} : memref<i32>
    %4 = affine.load %alloc[] : memref<i32>
    return %4 : i32
  }
  func.func @foo(%arg0: i32) -> i32 attributes {itypes = "s", otypes = "s"} {
    %0 = arith.extsi %arg0 : i32 to i33
    %c1_i32 = arith.constant 1 : i32
    %1 = arith.extsi %c1_i32 : i32 to i33
    %2 = arith.subi %0, %1 : i33
    %3 = arith.extsi %arg0 : i32 to i33
    %c1_i32_0 = arith.constant 1 : i32
    %4 = arith.extsi %c1_i32_0 : i32 to i33
    %5 = arith.addi %3, %4 : i33
    %6 = arith.divsi %2, %5 : i33
    %7 = arith.trunci %6 : i33 to i32
    %alloc = memref.alloc() {name = "B"} : memref<i32>
    affine.store %7, %alloc[] {to = "B"} : memref<i32>
    %8 = call @foo2(%arg0) : (i32) -> i32
    %9 = affine.load %alloc[] : memref<i32>
    %10 = arith.extsi %8 : i32 to i33
    %11 = arith.extsi %9 : i32 to i33
    %12 = arith.addi %10, %11 : i33
    %13 = arith.trunci %12 : i33 to i32
    %alloc_1 = memref.alloc() {name = "C"} : memref<i32>
    affine.store %13, %alloc_1[] {to = "C"} : memref<i32>
    %14 = affine.load %alloc_1[] : memref<i32>
    return %14 : i32
  }
  func.func @kernel(%arg0: i32) -> i32 attributes {itypes = "s", otypes = "s"} {
    %0 = arith.extsi %arg0 : i32 to i33
    %c1_i32 = arith.constant 1 : i32
    %1 = arith.extsi %c1_i32 : i32 to i33
    %2 = arith.addi %0, %1 : i33
    %3 = arith.trunci %2 : i33 to i32
    %alloc = memref.alloc() {name = "B"} : memref<i32>
    affine.store %3, %alloc[] {to = "B"} : memref<i32>
    %4 = affine.load %alloc[] : memref<i32>
    %5 = arith.extsi %arg0 : i32 to i64
    %6 = arith.extsi %4 : i32 to i64
    %7 = arith.muli %5, %6 : i64
    %8 = arith.trunci %7 : i64 to i32
    %alloc_0 = memref.alloc() {name = "C"} : memref<i32>
    affine.store %8, %alloc_0[] {to = "C"} : memref<i32>
    %9 = affine.load %alloc_0[] : memref<i32>
    %10 = arith.extsi %9 : i32 to i33
    %c1_i32_1 = arith.constant 1 : i32
    %11 = arith.extsi %c1_i32_1 : i32 to i33
    %12 = arith.addi %10, %11 : i33
    %13 = affine.load %alloc[] : memref<i32>
    %14 = arith.extsi %13 : i32 to i64
    %15 = arith.extsi %arg0 : i32 to i64
    %16 = arith.muli %14, %15 : i64
    %17 = arith.extsi %12 : i33 to i65
    %18 = arith.extsi %16 : i64 to i65
    %19 = arith.subi %17, %18 : i65
    %20 = arith.trunci %19 : i65 to i32
    %alloc_2 = memref.alloc() {name = "D"} : memref<i32>
    affine.store %20, %alloc_2[] {to = "D"} : memref<i32>
    %21 = affine.load %alloc_2[] : memref<i32>
    %22 = call @foo(%21) : (i32) -> i32
    %alloc_3 = memref.alloc() {name = "E"} : memref<i32>
    affine.store %22, %alloc_3[] {to = "E"} : memref<i32>
    %23 = affine.load %alloc_3[] : memref<i32>
    return %23 : i32
  }
}
```

Fangtangtang: The `assert` here has been changed back to `==`, but I’m not fully confident in [my fix](https://github.com/cornell-zhang/allo/blob/6facd02d34ebf1c51d4c1b3cd9d54ebc130e81b0/allo/passes.py#L697-L708).

(the fix is also explained below)

Fangtangtang:
**A major change**: for variables that have an init value at definition, their buffers are now directly taken from the rhs result. This should eliminate some redundant allocations and copies.

As we discussed earlier, I’ll need to double-check if the adjustments to the IR builder might hurt performance on existing tests/examples.
I wonder which test cases I should focus on, and which previous commit I should use as the baseline.

allo/passes.py
```diff
@@ -685,13 +694,17 @@ def add_use(val, val_name):
...
                 elif " = " in str(op):
-                    buf_name = f"{func_name}:{str(op).split(' = ', maxsplit=1)[0]}"
+                    if "from" in op.attributes:
+                        continue
+                    op_name = str(op).split(" = ", maxsplit=1)[0]
+                    buf_name = f"{func_name}:{op_name}"
+                    op.attributes["name"] = StringAttr.get(op_name, context=mod.context)
                 else:
-                    # call op does not have return value
+                    # op without return value
```
Fangtangtang, line 707: still using all op results as buffers here.
The result of load annotated with "from" is now treated as a use of the source, rather than defining a new buffer.

Fangtangtang:
For the systolic array, I can tell from the IR and `kernel.cpp` that there are now additional allocations and load/store operations, which could lead to performance regression. Based on the report, the performance does show a gap from the expected value (though that expected latency was likely an ideal target that couldn’t be achieved even in the original implementation). For 64x64x64 on 4x4 SA, the expected is 16384 but the actual result is 16898.
Since the [latest commit of the systolic array](https://github.com/cornell-zhang/allo/blob/21e87e1fd0317e102710a6a45424cc0c5860f474/tests/dataflow/test_multi_cache_gemm.py) isn’t runnable, it’s difficult for me to determine how severe the regression actually is. I wonder if there is a good way to retrieve the previous performance results.

chhzh123:
What's the error of the current SA?

Fangtangtang:
```txt
Error: Unsupported Name `packed_tmp`
╭──────────────────────────────────────────────────────────────────────── Traceback (most recent call last): Line: 123 ─────────────────────────────────────────────────────────────────────────╮
│  118:             packed_c: UInt(Rt * 8) = 0                                                                                                                                                  │
│  119:             for m in range(Rt):                                                                                                                                                         │
│  120:                 if m == i - 1:                                                                                                                                                          │
│  121:                     packed_c[m * 8:(m + 1) * 8] = c                                                                                                                                     │
│  122:                 else:                                                                                                                                                                   │
│  123:                     packed_c[m * 8:(m + 1) * 8] = packed_tmp[m * 8:(m + 1) * 8]                                                                                                         │
│  124:             L1_C[i - 1, j - 1].put(packed_c)                                                                                                                                            │
╰───────────────────────────────────────────────────────────────────────────────────────── Source Code ─────────────────────────────────────────────────────────────────────────────────────────╯
```
The root cause seems to be that this test case was developed on an old builder version, but when it got merged, the IR builder had already changed the scoping logic.

I could copy the test case back to a version before changing scoping, but I guess other logic in that version might also affect the results or performance. I think It would be better to have the commit hash of the version where this test case was developed, or at least the optimal performance numbers from that version.

PR #450, opened 2025-10-31
[Feature][Backend][VitisHLS] Add wide-bitwidth support in `vitis_hls` backend

### LINKED ISSUE #448: [Feature] Packed 512b DRAM/HBM interface support
Hi. 
Could I request support for wide (512b) extenal memory interface in Allo kernels? I'd like to fetch/write 512b every cycle from DRAM/HBM to maximize the external memory bandwidth (e.g., load/store 16 float32s or 64 int8s at once).
This is related to the unresolved bit-slicing issue in https://github.com/cornell-zhang/allo/issues/334.

Attempts so far:

**Attempt 1: Using packed integer of Int(512)**

- Code: [test_packed_int512.py](https://github.com/user-attachments/files/23208996/test_packed_int512.py)
   def packed_add(input: Int(512)[1], output: Int(512)[1]):
       # unpack, add, and repack 16 float32 using "[i * 32:(i + 1) * 32]"

-  Result: Python couldn't handle 512b ints ("OverflowError: Python int too large to convert to C long")

**Attempt 2: Packed struct**

- Code: [test_packed_struct.py](https://github.com/user-attachments/files/23209003/test_packed_struct.py)
packed_type = Struct({f'v{i}': float32 for i in range(16)})
   def struct_add(input: packed_type[1], output: packed_type[1]):
       # unpack, add, and repack using "[f'v{i}']"

- Result: Allo didn't support array-of-struct types as external memory interface

Would appreciate it if you could consider adding related feature or suggest any workaround to enable packed wide external memory interfaces.

@chhzh123 @slbioig

Thanks!

Fangtangtang:
## Description ##
This PR fixes #448.

### Problems ###
For the `vitis_hls` backend, our previous support for wide-bitwidth data types was incomplete.
[Type conversions (ctype to allo type)](https://github.com/cornell-zhang/allo/issues/448#issuecomment-3466390244) for wide bitwidths were missing, and the original host code relied on reading data from text files with known data types. As a result, wide-bitwidth types not directly supported by standard C++ could not be transferred correctly.

### Proposed Solutions ###
This PR adds the missing data type conversion logic and updates the host code to perform data transmission in a byte stream format.

### Examples ###
A valid example using `int512` (`Int(512)`)
```python3
    np_512 = get_np_struct_type(512)

    def packed_add(input_: int512[1], output_: int512[1]):
        unpacked: int32[16]
        packed: int512

        for i in range(16):
            unpacked[i] = input_[0][i * 32 : (i + 1) * 32]

        for i in range(16):
            unpacked[i] = unpacked[i] + 1

        for i in range(16):
            packed[i * 32 : (i + 1) * 32] = unpacked[i]

        output_[0] = packed

    A = np.random.randint(0, 64, (512 // 32,)).astype(np.int32)
    B = np.zeros(512 // 32).astype(np.int32)
    packed_A = np.ascontiguousarray(A).view(np_512)
    packed_B = np.ascontiguousarray(B).view(np_512)
    s = allo.customize(packed_add)
    if hls.is_available("vitis_hls"):
        mod = s.build(
            target="vitis_hls", mode="sw_emu", project="test_packed_int512.prj"
        )
        mod(packed_A, packed_B)
        unpacked_B = packed_B.view(np.int32)
        np.testing.assert_allclose(A + 1, unpacked_B, rtol=1e-5, atol=1e-5)
        print(unpacked_B)
        print("Passed Test!")
```

You can find more examples in `tests/test_vitis.py`.

## Checklist ##

- [ ] Code is well-documented

### DIFF
--- a/tests/test_vitis.py
+++ b/tests/test_vitis.py
@@ -130,10 +130,11 @@ def test_scalar():
     assert mod(1) == 2
     print("Passed CPU simulation!")
     mod = s.build(target="vitis_hls", mode="sw_emu", project="test_scalar_sw_emu.prj")
-    ret = np.zeros((1,), dtype=np.int32)
-    mod(1, ret)
-    assert ret == 2
-    print("Passed!")
+    if hls.is_available("vitis_hls"):
+        ret = np.zeros((1,), dtype=np.int32)
+        mod(1, ret)
+        assert ret == 2
+        print("Passed!")
 
 
 def test_pointer_generation():


allo/backend/hls.py
```diff
@@ -281,6 +296,7 @@ def __init__(
                 )
                 self.args = []
                 generate_makefile(dst_path, project, self.platform)
+                # [NOTE] (Shihan): I guess tapa backend do not use this one. I modified codegen_host for vitis, similar logic should be updated for tapa if self.host_code is useful here
                 self.host_code = codegen_host(
```
Fangtangtang, line 300: The modified `codegen_host` is also used in `tapa` backend, but I haven't updated corresponding parts for  `tapa`.
seems like the result `self.host_code` is not used for `tapa` backend, so I think this change should not break anything.

Please let me know if I misunderstood anything.

Fangtangtang:
A small suggestion:

Many of our backends pass parameters via file I/O (between the allo interface and the actual C++ host call). It seems that our existing parameter passing mostly uses text files. While text files are more readable and easier to debug, they can sometimes lead to unexpected errors.
I think using binary files (raw byte streams) may be a better approach for consistency and correctness. I’ve already switched the `aie` backend and the `vitis_hls` backend (this PR) to use binary mode, but a few other backends likely have the same issue.

It might be worth creating a **unified interface** for argument passing across all backends, so that this change can be applied consistently everywhere.

allo/backend/vitis.py
```diff
@@ -90,85 +88,57 @@ def codegen_host(top, module):
...
-                )
-            else:
-                out_str += format_str(f"ifile{i} >> source_in{i};")
-        else:
+            out_str += format_str(f"std::vector<uint8_t>  source_in{i}_({byte_num});")
             out_str += format_str(
-                f"{in_dtype} in_data_{i}[{'*'.join(map(str, in_shape))}];"
+                f"ifile{i}.read(reinterpret_cast<char*>(source_in{i}_.data()), {byte_num});"
             )
+            scalar_type = ctype_map[in_dtype]
```
Fangtangtang, line 125: Due to certain type consistency issues, it seems unavoidable to read scalars on the host side according to their data type. As a result, scalars can only be safely used in cases where there is a corresponding C++ type?

(If so, we'd better document this somewhere)

PR #453, opened 2025-11-05
[Frontend][IR][Builder][Simulator] Unroll gather/scatter in frontend for simulation

Oh-Afterglow:
## Description ##
This PR adds simulator support for `gather`/`scatter` (#432) by unrolling the operations in frontend, as well as trying to replace `meta_for`s in FEATHER with `gather`/`scatter`s.

(More test cases are needed other than test_collective_communication.py. And it seems that `gather`/`scatter`s can't simplify describing the butterfly network of FEATHER very much.)

(Edit: Draft PR -> Open PR. Removed the dummy `scf.execute_region` operation.)

### Problems ###
In PR #432 no unrolling will be done for `gather`/`scatter`. Only one FIFO will be generated for one such operation. For example, 
https://github.com/cornell-zhang/allo/blob/29fcfbf2894f81155a6b0a087696441dcde508c0/tests/dataflow/aie/test_collective_communication.py#L30
will be built as
```mlir
%0 = allo.stream_construct() {loop_name = "S_gather_0", name = "pipe_0", stream_list = ["pipe_0", "pipe_1", "pipe_2", "pipe_3"], stream_symbolic_slice_list = ["0", "1", "2", "3"], symbolic_slice = "0"} : !allo.stream<memref<32xi16>, 2>
%alloc = memref.alloc() : memref<4x32xi16>
%c0 = arith.constant 0 : index
%c4 = arith.constant 4 : index
%c1 = arith.constant 1 : index
scf.for %arg1 = %c0 to %c4 step %c1 {
  %1 = allo.stream_get(%0, []) : !allo.stream<memref<32xi16>, 2> -> memref<32xi16>
  %subview_0 = memref.subview %alloc[%arg1, 0] [1, 32] [1, 1] : memref<4x32xi16> to memref<32xi16, strided<[1], offset: ?>>
  memref.copy %1, %subview_0 : memref<32xi16> to memref<32xi16, strided<[1], offset: ?>>
} {loop_name = "gather0", op_name = "S_gather_0"}
%subview = memref.subview %arg0[0, 0] [4, 32] [1, 1] : memref<4x32xi16> to memref<4x32xi16, strided<[32, 1]>>
memref.copy %alloc, %subview {to = "B"} : memref<4x32xi16> to memref<4x32xi16, strided<[32, 1]>>
```
where the loop accesses the same FIFO for each iteration, which doesn't comply with the required semantics for simulation.

### Proposed Solutions ###
Although there is an `extra_stream_info` for AIE backend, building an unrolled loop in the frontend directly will be more natural instead of doing transformation in the simulator backend after calling `move_stream_to_interface`. Now if `unrolling=True`, all FIFOs will be generated for `gather`/`scatter`s, and the loop will be unrolled for accessing different FIFOs in different iterations. With unrolling, extra transformation is not needed to simulate `gather`/`scatter`s.

### Examples ###
The above example now builds to the following code if `unrolling=True`:
```mlir
%0 = allo.stream_construct() {name = "pipe_0"} : !allo.stream<memref<32xi16>, 2>
%1 = allo.stream_construct() {name = "pipe_1"} : !allo.stream<memref<32xi16>, 2>
%2 = allo.stream_construct() {name = "pipe_2"} : !allo.stream<memref<32xi16>, 2>
%3 = allo.stream_construct() {name = "pipe_3"} : !allo.stream<memref<32xi16>, 2>
%alloc = memref.alloc() : memref<4x32xi16>
%c0 = arith.constant 0 : index
%c1 = arith.constant 1 : index
%c2 = arith.constant 2 : index
%c3 = arith.constant 3 : index
%4 = allo.stream_get(%0, []) : !allo.stream<memref<32xi16>, 2> -> memref<32xi16>
%subview_0 = memref.subview %alloc[%c0, 0] [1, 32] [1, 1] : memref<4x32xi16> to memref<32xi16, strided<[1], offset: ?>>
memref.copy %4, %subview_0 : memref<32xi16> to memref<32xi16, strided<[1], offset: ?>>
%5 = allo.stream_get(%1, []) : !allo.stream<memref<32xi16>, 2> -> memref<32xi16>
%subview_1 = memref.subview %alloc[%c1, 0] [1, 32] [1, 1] : memref<4x32xi16> to memref<32xi16, strided<[1], offset: ?>>
memref.copy %5, %subview_1 : memref<32xi16> to memref<32xi16, strided<[1], offset: ?>>
%6 = allo.stream_get(%2, []) : !allo.stream<memref<32xi16>, 2> -> memref<32xi16>
%subview_2 = memref.subview %alloc[%c2, 0] [1, 32] [1, 1] : memref<4x32xi16> to memref<32xi16, strided<[1], offset: ?>>
memref.copy %6, %subview_2 : memref<32xi16> to memref<32xi16, strided<[1], offset: ?>>
%7 = allo.stream_get(%3, []) : !allo.stream<memref<32xi16>, 2> -> memref<32xi16>
%subview_3 = memref.subview %alloc[%c3, 0] [1, 32] [1, 1] : memref<4x32xi16> to memref<32xi16, strided<[1], offset: ?>>
memref.copy %7, %subview_3 : memref<32xi16> to memref<32xi16, strided<[1], offset: ?>>
%subview = memref.subview %arg0[0, 0] [4, 32] [1, 1] : memref<4x32xi16> to memref<4x32xi16, strided<[32, 1]>>
memref.copy %alloc, %subview {to = "B"} : memref<4x32xi16> to memref<4x32xi16, strided<[32, 1]>>
```

## Checklist ##

- [x] Code is well-documented

### DIFF
--- a/allo/ir/builder.py
+++ b/allo/ir/builder.py
@@ -13,7 +13,6 @@ import numpy as np
 from .._mlir.ir import (
     OpView,
     Module,
-    Block,
     Location,
     InsertionPoint,
     FunctionType,


chhzh123:
@Fangtangtang Does the unrolling impact the AIE backend?

Fangtangtang:
> Does the unrolling impact the AIE backend?

`ctx.unroll` is used as a conditional control here. The AIE backend has `ctx.unroll=False` by default, so it won’t be affected.

chhzh123:
ok, then can you help review this PR? @Fangtangtang

Fangtangtang:
ok

allo/ir/builder.py
```diff
@@ -2420,26 +2426,44 @@ class ExternalModule:
...
+                        for i in range(len(stream_ops)):
+                            index_const_expr = arith_d.ConstantOp(
+                                arith_d.IndexType.get(), i, ip=ctx.get_ip()
+                            )
+                            indices.append(index_const_expr)
+                        exec_region_op = scf_d.ExecuteRegionOp([], ip=ctx.get_ip())
+                        op_name = f"S_{fn_name}"
+                        exec_region_op.attributes["op_name"] = StringAttr.get(op_name)
+                        block = Block.create_at_start(exec_region_op.region, [])
+                        scf_d.YieldOp([], ip=InsertionPoint(block))
```
Fangtangtang: I'm curious why `scf.execute_region` is used here. It seems we could simply append the operations with `ip=ctx.get_ip()` instead.
if it's only for returning a value for `"scatter"`, it should be totally fine to just return `None`.

mlir/lib/Transforms/MemRefDCE.cpp
```diff
@@ -58,6 +58,12 @@ void removeNeverLoadedMemRef(func::FuncOp &func) {
       } else if (isa<StructConstructOp>(u)) {
         loaded_from = true;
         break;
+      } else if (isa<memref::CopyOp>(u)) {
+        loaded_from = true;
+        break;
+      } else if (isa<memref::ReinterpretCastOp>(u)) {
+        loaded_from = true;
+        break;
```
Fangtangtang, line 66: Thanks for adding the missing ones here! 
Feels a bit fragile to hardcode these cases. I know this isn’t an issue introduced in this PR, but I’m a bit worried we might overlook other similar cases.

Fangtangtang:
Also, I think we’ll need to further distinguish some control flags in the IR builder later.
`ctx.unroll` used in this PR was originally intended to control whether `df.kernel` instances are fully constructed (sorry, the naming is indeed confusing). Using `ctx.unroll` here works fine since currently only the AIE backend sets `ctx.unroll = False`.
We do have another flag (the environment variable `FORCE_UNROLL_INDEX`) that controls `meta_for` unrolling, which might be more suitable in this case.

Totally fine to use `ctx.unroll` in this PR. I'm just noting that we should probably standardize and document these flags later.

allo/ir/builder.py
Oh-Afterglow: Thanks for pointing out this. Additionally to avoiding potential problems caused by `None`, this operation is also used to preserve the attribute showing the code is from a `gather`/`scatter`, like the `for_op`. But this actually doesn't matter to the simulator backend. Maybe it'll be better to still return `None` here, since `build_Call` also returns nothing on other branches.

mlir/lib/Transforms/MemRefDCE.cpp
Oh-Afterglow: Yes, I agree with you. 
It seems that this `removeNeverLoadedMemRef` function is mainly used for removing `memref`s with type `memref<allo.struct>`, thus also removing the stored `allo.StructConstructOp`s, after lowering structs. Maybe we can just target such special `memref`s and let the existing DCE pass do the rest of work automatically for us.
I think the function `cleanUpUnusedOps` in the same file is also a bit problematic. It doesn't consider some side effects. I noticed that when it incorrectly removed some async dialect operations. (I tried to fix this earlier but haven't created a PR, since the async-related code is not prepared to be merged yet.)

allo/ir/builder.py
```diff
@@ -2466,37 +2483,46 @@ class ExternalModule:
...
+                                    static_strides=stride_values,
+                                    offsets=offset_values,
+                                    sizes=[],
+                                    strides=[],
+                                )
+                                allo_d.StreamPutOp(stream_op.result, [], subview.result)
                     if fn_name == "gather":
                         return alloc_op
-                    return for_op
+                    return for_op if not ctx.unroll else None
```
Fangtangtang, line 2525: also totally fine to return `None` when `ctx.unroll == True`, `for_op` is just a placeholder since the return value is never used elsewhere😅

chhzh123 (approved):
LGTM, thanks!

PR #482, opened 2025-12-11
[Backend][HLS] Emit linearized indices for function argument memrefs in Vivado

EzraReiss:
… HLS translation to correctly handle pointers.

## Description ##

Fix Vivado HLS code generation to use linearized pointer indexing for function argument memrefs instead of invalid multi-dimensional array syntax.

### Problems ###

When Allo translates multi-dimensional memref function arguments to Vivado HLS C++, the generated code uses multi-dimensional array syntax (e.g., `ptr[i][j][k]`) for accessing these arrays. However, function argument memrefs are actually passed as **pointers** in HLS (e.g., `T* ptr`), not as true multi-dimensional arrays. This causes compilation errors or incorrect behavior because pointer arithmetic requires **linearized indexing** (e.g., `ptr[i * D1 * D2 + j * D2 + k]`).

### Proposed Solutions ###

This PR modifies `EmitVivadoHLS.cpp` to:

1. **Detect function argument memrefs** using a new helper `isFunctionArgument()` that checks if a value originates from a `BlockArgument` of a `func.func` operation.

2. **Emit linearized indices for function arguments**: For `affine.load`, `affine.store`, `memref.load`, and `memref.store` operations targeting function argument memrefs, the translation now computes and emits the linearized (row-major) index instead of multi-dimensional bracket syntax.

3. **Preserve existing behavior for local arrays**: Local arrays declared within the function continue to use multi-dimensional indexing, which is correct for stack-allocated C arrays.

### Examples ###

**Before (incorrect for pointers):**
```cpp
void kernel(int8_t* Q) {
    val = Q[i][j][k];  // Error: Q is a pointer, not a 3D array
}

### DIFF
--- a/tests/test_vhls.py
+++ b/tests/test_vhls.py
@@ -281,6 +281,69 @@ def test_wrap_nonvoid(flatten):
     print("Passed!")
 
 
+def test_wrap_io_linearized_index():
+    M, N = 4, 4
+
+    def matrix_copy(A: int32[M, N]) -> int32[M, N]:
+        B: int32[M, N]
+        for i, j in allo.grid(M, N, name="copy"):
+            B[i, j] = A[i, j]
+        return B
+
+    s = allo.customize(matrix_copy)
+
+    # Test wrap_io=True
+    with tempfile.TemporaryDirectory() as tmpdir:
+        mod = s.build(target="vitis_hls", mode="sw_emu", project=tmpdir, wrap_io=True)
+        hls_code = mod.hls_code
+
+        # Check that function arguments are declared as pointers
+        assert (
+            "int32_t *v" in hls_code
+        ), "Expected pointer declaration for function arguments with wrap_io=True"
+
+        # Check that helper functions (load_buf/store_res) exist for data movement
+        assert (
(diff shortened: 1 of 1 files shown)


chhzh123:
@EzraReiss We just upgraded the LLVM version, so probably you are not able to test correctness on our server or CI right now. Could you rebase to the latest commit and resolve the formatting and linting issues first?

EzraReiss:
im trying to do this rn

EzraReiss:
> @EzraReiss We just upgraded the LLVM version, so probably you are not able to test correctness on our server or CI right now. Could you rebase to the latest commit and resolve the formatting and linting issues first?

I think were good it built

chhzh123 (approved):
Awesome! This looks fantastic. One minor issue: could you add a test case under `tests/test_vhls.py` to check that the generated code is correct?

EzraReiss:
> Awesome! This looks fantastic. One minor issue: could you add a test case under `tests/test_vhls.py` to check that the generated code is correct?

I tested it myself with an hbm kernel for the QKV projection in attention but am writing an official test case now

PR #487, opened 2025-12-13
[Feature][IR] Add stateful type qualifier to Allo for persistent state

maheshejs:
## Description ##

This pull request adds first-class support for stateful variables in the Allo programming model. It allows hardware accelerators written in Allo's Python-based DSL to express persistent state (e.g., accumulators, FSMs, or scratchpad memories) that is automatically preserved across kernel invocations, much like C++ static variables or hardware registers. This feature enables more natural and efficient modeling of a wide range of accelerator designs, improving both productivity and performance.

### Problems ###

Previously, Allo only supported stateless kernels: all local variables are re-initialized for each kernel call, and there was no direct way to retain data across invocations. This limited expressiveness, making it hard or inefficient to model hardware modules like accumulators, shift registers, or persistent scratchpads. Common workarounds (e.g., passing state in/out via kernel arguments or the host) are cumbersome, error-prone, and result in substantial data movement overhead.

### Proposed Solutions ###

This PR adds a stateful type qualifier to the Allo Python frontend and propagates this feature through the compilation flow:

- *Frontend Syntax*: You can now declare stateful variables using stateful(type) in kernel bodies, e.g., acc: stateful(int32) = 0. This marks acc as persistent state.
- *Type System*: Adds a .stateful attribute to Allo types (AlloType), and strictly prevents stateful variables from being used as kernel arguments.
- *MLIR Generation*: Stateful variables are lowered to global memory (memref.global) in MLIR, initialized once and shared by all calls to a particular kernel instance.
- *Backend Codegen*: Emits function-local static variables in HLS C++ (Vivado/Vitis HLS) for all stateful variables—ensuring their persistence and synthesizability.
- *Instance Isolation*: When a kernel with stateful variables is instantiated multiple times, each instance gets its own independent state (via symbol mangling and MLIR rewriting). This enables compositional hardware designs where each accelerator block has its isolated persistent state.

All changes are compatible with existing stateless programs.

### Examples ###

Example 1: Stateful Scalar Accumulator

Allo Python:

```python
from allo.ir.types import int32, stateful

def accumulator(x: int32) -> int32:
    state: stateful(int32) = 0
    state = state + x
    return state
```

Lowered MLIR:

```mlir
module {
  memref.global "private" @__stateful_accumulator_state_1 : memref<i32> = dense<0>
  func.func @accumulator(%arg0: i32) -> i32 attributes {itypes = "s", otypes = "s"} {
    %0 = memref.get_global @__stateful_accumulator_state_1 : memref<i32>
    %1 = affine.load %0[] {from = "state"} : memref<i32>
    %2 = arith.extsi %1 : i32 to i33
    %3 = arith.extsi %arg0 : i32 to i33
    %4 = arith.addi %2, %3 : i33
    %5 = arith.trunci %4 : i33 to i32
    affine.store %5, %0[] {to = "state"} : memref<i32>
    %6 = affine.load %0[] {from = "state"} : memref<i32>
    return %6 : i32
  }
}
```

Generated HLS C++:

```cpp
void accumulator(
  int32_t v0,
  int32_t *v1
) {
  static int32_t __stateful_accumulator_state_1 = {0};
  int32_t v3 = __stateful_accumulator_state_1;
  int32_t v4 = v0;
  int32_t v5 = v3 + v4;
  __stateful_accumulator_state_1 = v5;
  *v1 = __stateful_accumulator_state_1;
}
```

Example 2: Stateful CRC32 Calculation

Allo Python:

```python
from allo.ir.types import uint8, uint32, stateful

def update_crc32(data: uint8) -> uint32:
    crc: stateful(uint32) = 0xFFFFFFFF
    crc = crc ^ data
    for j in range(8):
        crc = (crc >> 1) ^ (0xEDB88320 & (-(crc & 1)))
    return crc ^ 0xFFFFFFFF
```

Relevant generated HLS C++:

```cpp
void update_crc32(
  uint8_t v0,
  uint32_t *v1
) {
  static int32_t __stateful_update_crc32_crc_1 = {-1};
  int32_t v3 = __stateful_update_crc32_crc_1;
  uint32_t v4 = v0;
  uint32_t v5 = v3 ^ v4;
  __stateful_update_crc32_crc_1 = v5;
  for (int j = 0; j < 8; j++) {
    int32_t v7 = __stateful_update_crc32_crc_1;
    uint32_t v8 = v7 >> 1;
    int32_t v9 = v7 & 1;
    int32_t v10 = 0 - v9;
    int32_t v11 = v10 & -306674912;
    int32_t v12 = v8 ^ v11;
    __stateful_update_crc32_crc_1 = v12;
  }
  int32_t v13 = __stateful_update_crc32_crc_1;
  *v1 = v13 ^ -1;
}
```

On each invocation, crc will persist across calls as in hardware, which accurately models shift registers or accumulators.

## Checklist ##

- [x] Code is well-documented

## Contributors ##

Jenny Lee
Jifeng Wu
Joseph Maheshe @maheshejs 
Sunwoo Kim @sunwookim028

### DIFF
--- a/allo/customize.py
+++ b/allo/customize.py
@@ -1162,7 +1162,7 @@ class Schedule:
             if not isinstance(sch, Schedule):
                 raise TypeError("The first argument must be a Schedule object")
 
-            if hasattr(sch, "stateful_var_map"):
+            if hasattr(sch, "stateful_var_map") and sch.stateful_var_map:
                 stateful_seen = getattr(self, "_stateful_seen", {})
                 self._stateful_seen = stateful_seen
 


chhzh123:
Hey @maheshejs, we just upgrade the LLVM version, so if you are using our server, maybe you need to wait for our update before you can use the latest commit

chhzh123:
@maheshejs @sunwookim028 Can you guys rebase your code to the latest commit? Also try to separate the mini TPU implementation from this PR

chhzh123:
@maheshejs @sunwookim028 I've upgraded the LLVM on brg-zhang. Please follow the guide below to set up the new environment
```bash
# unset the original setup-llvm-19 script
source /work/shared/common/allo/setup-llvm-main.sh
# pull the latest commit
cd allo
git pull
# install additional packages and reinstall Allo
python3 -m pip install pybind11 nanobind
export PYBIND11_CMAKE_DIR="$(python3 -c 'import pybind11; print(pybind11.get_cmake_dir())')"
export NANOBIND_CMAKE_DIR="$(python3 -c 'import nanobind; print(nanobind.cmake_dir())')"
python3 -m pip install -v -e .
```

chhzh123:
@maheshejs Can you do a rebase instead of a merge so we don't see so many commits?

maheshejs:
> @maheshejs Can you do a rebase instead of a merge so we don't see so many commits?

i see, there were 300+ commits. the main branch we started with diverged a while ago and the merge just made it messy. i just resetted our branch on top of upstream and included changes we did, which are only a handful of files for a single commit. hopefully, that will do...

chhzh123:
Sure, if that's easier, you can create another PR

maheshejs:
> Sure, if that's easier, you can create another PR

i did it in this PR

chhzh123:
Ah I saw your latest commit. Just let me know if this PR is ready for review

chhzh123:
Could you also cc your collaborators in the description if they helped implement this feature? and turn the draft PR into "Ready for review" status

maheshejs:
> Could you also cc your collaborators in the description if they helped implement this feature? and turn the draft PR into "Ready for review" status

Added collaborators section in the description and made the PR ready for review.

chhzh123:
Do you plan to change the syntax in the next PR? Per offline discussion, we want to change the `Stateful` type as the attribute of the base type, as shown below

```python
from allo.ir.types import int32, Stateful
state: int32 @ Stateful
```

maheshejs:
> Do you plan to change the syntax in the next PR? Per offline discussion, we want to change the `Stateful` type as the attribute of the base type, as shown below
> 
> ```python
> from allo.ir.types import int32, Stateful
> state: int32 @ Stateful
> ```

Yes, planning to do that, to `state: int32 @ {Stateful}` in a list for future attributes

allo/ir/builder.py
```diff
@@ -1727,6 +1757,160 @@ def build_Dict(ctx: ASTContext, node: ast.Dict):
...
+                    Parameters
+                    ----------
+                    allo_type : AlloType
+                        The Allo type to convert
+
+                    Returns
+                    -------
+                    numpy dtype
+                        Corresponding numpy data type
+                    """
```
chhzh123: Can you put this function under `allo/utils.py`?

examples/stateful/accumulator.py
chhzh123, line 1: Can you move these small examples to a test file called `test_stateful.py` under the `tests` folder? Please add the necessary pytest facilities to test correctness

examples/stateful/drive_tinytpu.cpp
chhzh123, line 1: Remove. TinyTPU should be in another PR

examples/stateful/test_shihans_vpu.py
chhzh123, line 1: Remove

examples/stateful/test_stateful.py
chhzh123, line 1: Add pytest and assertions. Need to actually build and run using the LLVM backend

examples/stateful/test_stateful_hls.py
chhzh123, line 1: For HLS testing, test if the generated code structure and syntax are expected. For example, checking the `static` qualifier.

examples/stateful/test_stateful_isolate.py
chhzh123, line 1: Can be merged with the above HLS file

examples/stateful/test_stateful_kernel_rng.py
chhzh123, line 1: Same, merge into one large test file

examples/stateful/test_tiled_systolic_stateful.py
```diff
@@ -0,0 +1,116 @@
...
+                    # drain
+                    with allo.meta_elif(i == Mt + 1):
+                        for kt in range(Kt):
+                            b: int32 = fifo_B[i, j].get()
+                    with allo.meta_elif(j == Nt + 1):
+                        for kt in range(Kt):
+                            a: int32 = fifo_A[i, j].get()
+                    # main body
+                    with allo.meta_else():
+                        c: stateful(int32) = 0
```
chhzh123: Why do we need `c` to be stateful in this design? This design makes all the tiling loops inside the kernel

examples/stateful/test_tinyalu.py
```diff
@@ -0,0 +1,44 @@
...
+    if op == OP_MUL:
+        mem[addr] = int32_mul(mem[addr], mem[addr + 1])
+        retval = mem[addr]
+    return retval
+
+
+s = allo.customize(test_tpu)
+# x = s.build(target="llvm")
+
+mod = s.build(target="vitis_hls", mode="sw_emu", project="/tmp/tpu.prj")
```
chhzh123: Use the LLVM backend for software testing

examples/stateful/test_tinytpu.py
chhzh123, line 1: Again, move the TPU-related files to the next PR

mlir/lib/Translation/EmitVivadoHLS.cpp
```diff
@@ -423,7 +424,7 @@ class AffineExprEmitter : public AlloEmitterBase,
       }
     }
     if (auto binaryRHS = llvm::dyn_cast<AffineBinaryOpExpr>(expr.getRHS())) {
-      if (auto constRHS =
+      if (auto constRHS =
```
chhzh123: Need to re-format this with clang-format

mlir/lib/Translation/EmitVivadoHLS.cpp
```diff
@@ -1477,6 +1478,18 @@ void ModuleEmitter::emitGlobal(memref::GlobalOp op) {
     indent();
     auto arrayType = llvm::dyn_cast<ShapedType>(op.getType());
     auto type = arrayType.getElementType();
+    // Check for hls.static attribute or stateful variable naming pattern
+    bool isStatic = op->hasAttr("hls.static");
```
chhzh123: Did I miss anything? I didn't see any references for `hls.static` in the builder code

chhzh123: Also, don't use a target-specific attribute for high-level IR. You can just annotate it as `static`

allo/ir/builder.py
```diff
@@ -157,7 +157,22 @@ def build_assign_value(ctx: ASTContext, node: ast.Name, buffer, val):
     @staticmethod
     def build_Name(ctx: ASTContext, node: ast.Name, val=None):
         if val is not None and isinstance(node.ctx, ast.Store):
-            buffer = ctx.get_symbol(node.id)
+            if hasattr(ctx, "custom_globals") and node.id in ctx.custom_globals:
```
chhzh123: Why not just reuse `ctx.global_vars`?

maheshejs:
@sunwookim028  please don't merge, it adds many commits and messes the history

allo/ir/builder.py
maheshejs: they have different metadata, and `custom_globals` specifically tracks stateful variables. for composition isolation in `customize`, for instance, we need to go through stateful variables and not all the `global_vars` which include functions too and other metadata...

allo/ir/builder.py
maheshejs: Done.

mlir/lib/Translation/EmitVivadoHLS.cpp
maheshejs: Used `static`.

chhzh123:
@maheshejs Can you fix the pylint issues?

chhzh123:
also let me know if it is ready to review

examples/stateful/test_tiled_systolic_stateful.py
maheshejs: It was to tile the innermost dimension too, `k`. However, yes, stateful is not really needed since that innermost loop could just be tiled like the other ones. I removed this test for now and plan to merge with a more valid one using the df.kernel.

maheshejs:
> @maheshejs Can you fix the pylint issues?

I addressed all of them, except too-many-instance-attributes since I added 2 more on top of the existing 11. I'm wondering how to address that. That aside, it's ready to review.

chhzh123:
If you think it is fine for your case, you can add `# pylint: disable=too-many-instance-attributes` to skip that check

allo/ir/builder.py
```diff
@@ -159,7 +160,22 @@ def build_assign_value(ctx: ASTContext, node: ast.Name, buffer, val):
...
+                    buffer = ctx.func_globals[global_name]
+                else:
+                    # Create new op and cache it
+                    memref_type = ctx.custom_globals[node.id][1]
+                    get_global_op = memref_d.GetGlobalOp(
+                        memref_type,
+                        FlatSymbolRefAttr.get(global_name),
+                        ip=ctx.get_ip(),
+                    )
+                    ctx.func_globals[global_name] = get_global_op
```
chhzh123: Maybe give better names for `func_globals` and `custom_globals`. Not super straightforward to see what they store from the names

allo/ir/builder.py
```diff
@@ -1826,6 +1943,8 @@ def build_AnnAssign(ctx: ASTContext, node: ast.AnnAssign):
 
     @staticmethod
     def build_FunctionDef(ctx: ASTContext, node: ast.FunctionDef):
+        if not hasattr(ctx, "func_globals"):
+            ctx.func_globals = {}
```
chhzh123: What will happen if we have nested function calls and both functions have the same stateful variable name?

allo/ir/types.py
```diff
@@ -76,15 +77,38 @@ def __getitem__(self, sizes):
...
             return False
-        return self.name == other.name
+        return self.name == other.name and self.stateful == other.stateful
 
     def __hash__(self):
-        return hash(self.name)
+        return hash((self.name, self.stateful))
+
+
+def stateful(dtype: AlloType):
```
chhzh123, line 93: Mark a TODO here to denote the expected type declaration

mlir/lib/Translation/EmitVivadoHLS.cpp
```diff
@@ -3098,4 +3235,4 @@ void allo::registerEmitVivadoHLSTranslation() {
         >();
         // clang-format on
       });
-}
\ No newline at end of file
+}
```
chhzh123, line 3078: Empty line in the end

allo/ir/builder.py
maheshejs: Yup, right. This seems to be the case of the failing CI tests. Looking into it.

mlir/lib/Translation/EmitVivadoHLS.cpp
maheshejs: add empty line or deleting? adding doesn't pass clang-format.

allo/ir/builder.py
maheshejs: Actually, not related. For that case, there won't be a collision since function names are embedded in the stateful variable name. Adding a test for that.

maheshejs:
> Mostly LGTM. Can you also add more complicated examples with multiple functions in the test cases? For example, the stateful variable is not in the top-level function, and there are multiple stateful variables in different subfunctions.

Good, I added two tests with multiple functions/subfunctions. CI tests are passing, but Pytorch setup is not and I believe our changes shouldn't involve Pytorch. Any ideas?

chhzh123:
Did your changes modify the function naming? Probably you need to print out the Allo/MLIR modules for the failed cases to see what happened.

maheshejs:
> Did your changes modify the function naming? Probably you need to print out the Allo/MLIR modules for the failed cases to see what happened.

Fixed, something related to the changes in customized. CI is passing now. It's ready for review too.

allo/customize.py
```diff
@@ -1161,6 +1161,49 @@ def get_name(arg):
...
+                        (
+                            op
+                            for op in self.module.body.operations
+                            if isinstance(op, memref_d.GlobalOp)
+                            and op.attributes["sym_name"].value == global_name
+                        ),
+                        None,
+                    )
+                    if original_global is None:
+                        raise RuntimeError(f"Stateful global {global_name} not found")
```
Fangtangtang, line 1188: I think there is an issue with how stateful variable names are assigned (in builder) and rewritten.

Taking `tests/test_stateful_hls.py::test_stateful_isolate_hls` as an example: both stateful variables in `s` are named `__stateful_acc_state_1`. During
```python
s.compose(s1, id="A1")
s.compose(s2, id="A2")
```
the module is scanned for stateful operations, and variables named `__stateful_acc_state_1` are rewritten sequentially.

However, the intended behavior seems to be that `s1` is composed precisely as `acc[int32, "A1"](x)`. With the current name-based replacement, swapping the order of these two scheduling steps can lead to unexpected results?

---

I'm trying to do some cleanup, so please correct me if my understanding is wrong.

sunwookim028: @jw2858 Would you have a chance to check this?

PR #501, opened 2025-12-24
[CI][AIE] Enable CI for `tests/dataflow/aie` tests

Fangtangtang:
## Description ##
 Added a GitHub Actions workflow to run AIE-specific dataflow tests.

### Problems ###
AIE backend tests under were not included in CI.

### Proposed Solutions ###
Added a GitHub Actions workflow for AIE backend  tests.

Built a new Docker image for AIE CI to handle special NPU configurations required inside Docker. (reference dockerfile: https://github.com/AMDResearch/Ryzers/blob/main/packages/npu/xdna/Dockerfile)

## Checklist ##

- [ ] Code is well-documented

### DIFF
--- a/.github/workflows/aie_ci.yml
+++ b/.github/workflows/aie_ci.yml
@@ -1,19 +1,15 @@
 # Copyright Allo authors. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-name: "Allo Test"
+name: "Allo AIE Test"
 on:
-  push:
-    branches:
-      - main
-  pull_request:
-    branches:
-      - main
+  schedule:
+    - cron: '0 8 * * 0' # Sunday 08:00 UTC 
 
 jobs:
   aie_test:
     # This job runs on Linux
-    runs-on: [self-hosted, aie, xdna2]
+    runs-on: [zhang-ryzen2]
     container:
       image: shihanfang/allo-ci:test
       options: "--device /dev/accel/accel0:/dev/accel/accel0 --ulimit memlock=-1"
@@ -25,7 +21,7 @@ jobs:
         git config --global --add safe.directory "$GITHUB_WORKSPACE"
     - uses: actions/checkout@v4
       with:
-        ref: ${{ github.ref }}
+        ref: main 
         fetch-depth: 0
     - name: Build Allo
       shell: bash


Fangtangtang:
Using the NPU inside Docker still has the risk of test crashes that can leave the device in an abnormal state (and may require power off and restart to recover).

.github/workflows/aie_ci.yml
```diff
@@ -0,0 +1,60 @@
...
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  aie_test:
+    # This job runs on Linux
+    runs-on: [self-hosted, aie, xdna2]
```
chhzh123: Use `zhang-ryzen2` instead of these three?

.github/workflows/aie_ci.yml
```diff
@@ -0,0 +1,60 @@
+# Copyright Allo authors. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+name: "Allo Test"
```
chhzh123: Allo AIE Test

Fangtangtang:
> It seems the server is down.

right. 
And the runner is currently set up under my own home directory due to some read/write permission issues. I’m wondering how other runners are set up.

chhzh123:
Use `nohup` to keep the script running. I don't think who launches it really matters. Just avoid killing that process

Fangtangtang:
I'm also worried that PRs submitted at arbitrary times may trigger the CI and conflict with users who are actively using the NPU on the server, or even leave the NPU in an abnormal state during tests. I'm not sure keeping the it always online is a good solution. It might be better to configure the runner to be online only during specific time windows, or to enable it manually when needed.😓

chhzh123:
Can we set up a weekly action?

zhangzhiru:
Testing every week sounds like the right frequency.

Fangtangtang:
weekly on all the open PRs? Let me try.

chhzh123:
No need to be triggered by PR

Fangtangtang:
ok, that's easier

Fangtangtang:
@chhzh123 ready to merge. The schedule is set for Sunday 08:00 UTC. Please let me know if anything needs changing.

PR #518, opened 2026-01-12
[dataflow] Support hierarchical modules

chhzh123:
## Description ##
This PR adds support for hierarchical dataflow regions. Users can now define a submodule and then call it in another module. Each module follows the same SPMD programming style.

### Examples ###
```python
@df.region()
def inner(A: float32[M, K], B: float32[K, N], C: float32[M, N]):
    @df.kernel(mapping=[P0, P1], args=[A, B, C])
    def gemm(local_A: float32[M, K], local_B: float32[K, N], local_C: float32[M, N]):
        pi, pj = df.get_pid()
        for i in range(pi * Mt, (pi + 1) * Mt):
            for j in range(pj * Nt, (pj + 1) * Nt):
                for k in range(K):
                    local_C[i, j] += local_A[i, k] * local_B[k, j]


@df.region()
def top(A: float32[M, K], B: float32[K, N], C1: float32[M, N], C2: float32[M, N]):
    @df.kernel(mapping=[2], args=[A, B, C1, C2])
    def wrapper(
        local_A: float32[M, K],
        local_B: float32[K, N],
        local_C1: float32[M, N],
        local_C2: float32[M, N],
    ):
        i = df.get_pid()
        with allo.meta_if(i == 0):
            inner(local_A, local_B, local_C1)
        with allo.meta_if(i == 1):
            inner(local_A, local_B, local_C2)
```

## Checklist ##

Please make sure to review and check all of these items:
- [x] PR's title starts with a category (e.g. [Bugfix], [IR], [Builder], etc)
- [x] All changes have test coverage (It would be good to provide ~2 different test cases to test the robustness of your code)
- [x] Pass the [formatting check](https://cornell-zhang.github.io/allo/developer/index.html#id1) locally
- [x] Code is well-documented

PR #530, opened 2026-01-21
[agent] Add AGENTS.md

chhzh123:
## Description ##
As coding agents become more and more popular, this PR adds a ["constitution"](https://www.anthropic.com/news/claudes-constitution) `AGENTS.md` file to force all the agents working on this repository to follow some common rules. This file is supposed to act as a specification of the expected workflow and coding styles. It is built to be evolvable, and every time we encounter something that the agent cannot perform well, we should update this file or the related documentation to let it know what the right approach is to tackle the problem. For the configurations of different coding agents, please refer to the following links:

* Claude Code: https://www.anthropic.com/engineering/claude-code-best-practices
* Cursor: https://cursor.com/docs/context/rules#agentsmd
* Antigravity: https://antigravity.google/docs/rules-workflows

The current rules are listed below:
```markdown
# Building
- Run `pip install -v -e .` to build the full project (includes MLIR/C++ backend)
- Always `conda activate allo` before building or running tests
- Read `docs/source/dive/frontend_syntax.rst` for comprehensive Allo frontend syntax reference
- Read `docs/source/dive/dataflow.rst` for the dataflow programming model (regions, kernels, streams)

# Testing
- Run `bash scripts/lint/task_lint.sh` for formatting checks
- Run `python3 -m pytest --ignore=tests/dataflow tests -v` for tests
  - Prefer running a single test file instead of the full suite (full suite is slow)
  - Use only software simulators (`target="llvm"` or `target="simulator"`)
  - If Vitis HLS tests are needed, ask the user to run them manually

# Code style
- Make small, targeted diffs rather than large refactors
- Prefer general solutions instead of one-off `if/else` patches
- Place Python frontend code in `allo/`
- Place MLIR dialects and passes code in `mlir/`
- Add tests and documentation for new features in `tests/` and `docs/`

# Don'ts
- Do not modify repository structure without approval
- Do not install system packages without explicit user confirmation

```

cc: @Fangtangtang @zhangzhiru 

## Checklist ##

- [ ] Code is well-documented

### DIFF
--- a/docs/source/dive/dataflow.rst
+++ b/docs/source/dive/dataflow.rst
@@ -532,7 +532,7 @@ Import and Setup
 Applying Layouts to Kernel Arguments
 ------------------------------------
 
-Use the ``@`` operator to annotate kernel arguments with their layout:
+Use the ``@`` operator to annotate kernel arguments with their layout. If no layout is specified, the default is **replicated** for all dimensions:
 
 .. code-block:: python
 
@@ -541,7 +541,7 @@ Use the ``@`` operator to annotate kernel arguments with their layout:
     @df.kernel(mapping=[P0, P1], args=[A, B])
     def kernel(local_A: Ty[M, N] @ LyA, local_B: Ty[M, N]):
         # local_A is automatically partitioned according to LyA
-        # local_B has no layout annotation (uses defaults)
+        # local_B has no layout annotation (defaults to Replicate)
         pass
 
 Layout Patterns
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,6 +1,6 @@
 # Building
+- Always run `conda activate allo` before building or running tests
 - Run `pip install -v -e .` to build the full project (includes MLIR/C++ backend)
-- Always `conda activate allo` before building or running tests
 - Read `docs/source/dive/frontend_syntax.rst` for comprehensive Allo frontend syntax reference
 - Read `docs/source/dive/dataflow.rst` for the dataflow programming model (regions, kernels, streams)
 


Fangtangtang:
To avoid agent generated code deviating from our 'implicitly defined' language spec, I suggest explicitly formalize our language spec (syntax + semantics) and reference it in AGENTS.md.

chhzh123:
Yes! Updated the documentation for the frontend syntax and dataflow programming. This should serve as the "trusted base" or the contract between us and agents. @Fangtangtang Please check the details and comment if there is anything unclear.

docs/source/dive/dataflow.rst
```diff
@@ -0,0 +1,515 @@
...
+
+    @df.kernel(mapping=[P0, P1], args=[A, B, C])
+    def kernel_name(local_A: Ty[...], local_B: Ty[...], local_C: Ty[...]):
+        # Kernel body
+        pass
+
+**Parameters:**
+
+- ``mapping``: A list specifying the kernel's parallel instantiation shape
+- ``args``: A list of region-level arguments accessible by this kernel
```
Fangtangtang: It would be helpful to clarify the correspondence between the `args` list and the kernel arguments.
- The element at index `i` in the `args` list is passed as the `i`-th argument to the kernel function. (so the number and argument type must match)
- `args` is optional and can be omitted if the kernel takes no arguments.

docs/source/dive/frontend_syntax.rst
```diff
@@ -0,0 +1,922 @@
...
+
+Basic Function Signature
+------------------------
+
+Allo kernels are defined as Python functions with explicit type annotations for all arguments and return types.
+The syntax follows Python's type hint notation:
+
+.. code-block:: python
+
+    def kernel(arg1: Type1[Shape1], arg2: Type2[Shape2]) -> ReturnType[Shape]:
```
Fangtangtang: arguments and return types can be scalar. for,example
```python
def kernel(A: int32) -> int32:
   ...
```

docs/source/dive/frontend_syntax.rst
```diff
@@ -0,0 +1,922 @@
...
+    temp: int32[M, N] = 0
+    outp: int32[M, N] = temp  # Copy temp to outp
+
+    # Copy from argument
+    def kernel(inp: int32[M, N]) -> int32[M, N]:
+        outp: int32[M, N] = inp
+        return outp
+
+
+Loop Constructs
```
Fangtangtang, line 187: break or continue not supported?

docs/source/dive/frontend_syntax.rst
```diff
@@ -0,0 +1,922 @@
...
+Meta-Programming Constructs
+===========================
+
+Allo provides compile-time meta-programming constructs that are evaluated during compilation,
+enabling conditional code generation and advanced optimizations.
+
+Meta If/Elif/Else
+-----------------
+
+Compile-time conditionals that generate different code based on conditions known at compile time:
```
Fangtangtang: meta if/elif/else conditions and meta for arguments must be compile time constants

docs/source/dive/frontend_syntax.rst
```diff
@@ -0,0 +1,922 @@
...
+        with allo.meta_for(8) as i:
+            coef: ConstExpr[float32] = compute_coefficient(i)
+            B[i] = A[i] * coef
+
+.. note::
+
+   ConstExpr variables **must** be initialized at declaration time. Uninitialized ConstExpr will raise an error.
+
+
+Scoping Rules
```
Fangtangtang, line 614: It would be better to add definition for "scope boundaries" here. e.g.,
```md
Allo enforces C++-style **Block Scoping** rules, which differs from standard Python.

*   **Scope Boundaries**: `if`, `elif`, `else`, `for`, `while`, `meta_if`, `meta_for`, `meta_else`.
*   **Rule**: A variable declared for the first time inside a block is **local** to that block. It is not visible after the block exits.
*   **Access**: Inner blocks can read/write variables defined in outer blocks.
```

---

I also want to add **Reassignment Validity** somewhere.
*   A variable can be reassigned.
    *   The new value must match the declared type of the variable.
    *   **Immutable Constants**: `ConstExpr` variables and values returned by `df.get_pid()` are compile-time constants and **cannot** be reassigned.

Fangtangtang (commented):
Looks great overall! The only missing piece I noticed is the `Layout` type in dataflow programming.

Fangtangtang:
One more question: as we rely more on agents, development starts to resemble natural language programming, and I found 'intention communication' to be really fragile. agents may not fully capture the intent behind existing code, and it can also be hard to infer the intent behind agent generated changes.

Looking back at some of the changes made by agents, I noticed a few modifications that go against my original intention. I think this is largely because the agent didn't fully understand the intent behind the existing code (partly due to insufficient comments or documentation on my side). 
Conversely, I also find it hard to get the intent behind some agent generated code. For example, why `TypeInferer` is re-invoked inside the `builder` (introduced in #518). My intuition is that this could introduce redundant analysis, since the inference already happens before the builder stage.

I suppose the same will happen between agent A and agent B. just wondering if there is a good way for 'intention communication'

chhzh123:
> agents may not fully capture the intent behind existing code, and it can also be hard to infer the intent behind agent generated changes.

This is mostly because our code is not clear, either lacking comments or documentation, or the abstraction itself is not straightforward.

> I think this is largely because the agent didn't fully understand the intent behind the existing code (partly due to insufficient comments or documentation on my side).

Right, so we need to enhance our spec, make clear comments, and improve our documentation.

> Conversely, I also find it hard to get the intent behind some agent generated code.

This is an intention misalignment. It means the agent is malfunctioning. We should actually fix that.

> I suppose the same will happen between agent A and agent B. just wondering if there is a good way for 'intention communication'

The final decision is made by us. For the intention, we should be more explicit about what we want to achieve. Actually, if just making the code able to execute, the agent indeed can call any functions they want, similar to "reward hacking". In that case, we should add conciseness as the objective as well in the rule (as in this PR), and maybe create additional subrules saying how to handle the builder functions.

Moreover, we should enhance human involvement in the final decision. I'll tag you for review in the future PRs.

chhzh123:
@Fangtangtang Can you help check if there are missing pieces you want to mention in the spec?

Fangtangtang:
can we have the agent write tricky test cases based solely on its understanding of the spec? I think this could be a good way to both evaluate the spec and test our existing implementation.

Fangtangtang:
and, I'm still confused about how should we maintain our existing code base. Do we require developers to write docs for every module they have implemented, or should the agent generate docs based on its understanding of the code, with developers verifying them?
My concern is that if we let the agent develop from scratch and generate docs for each module, the systeconfum would be 'consistent' at some extend. But we are currently in a mixed mode, which feels like a chaotic system.

chhzh123:
No, nothing really needs to change right now. If anything is unclear, then open a new PR to improve the documentation or comments. This is not meant to be a one-time effort. We'll likely need a few iterations until it works well.

> we are currently in a mixed mode, which feels like a chaotic system.

Humans and agents will co-exist going forward, so most repositories will run into similar growing pains. From my perspective, an agent isn't fundamentally different from a human engineer--we can treat it like a human committer. We can teach an agent how to behave correctly the same way we onboard new students onto the codebase. The final decisions still rest with us, which is why PRs are a good mechanism to keep a human in the loop.

docs/source/dive/dataflow.rst
```diff
@@ -0,0 +1,685 @@
...
+Use the ``@`` operator to annotate kernel arguments with their layout:
+
+.. code-block:: python
+
+    LyA = [S(0), R]  # Shard first dim on mesh axis 0, replicate second dim
+
+    @df.kernel(mapping=[P0, P1], args=[A, B])
+    def kernel(local_A: Ty[M, N] @ LyA, local_B: Ty[M, N]):
+        # local_A is automatically partitioned according to LyA
+        # local_B has no layout annotation (uses defaults)
```
Fangtangtang: shall we add "default to replicate"?

docs/source/dive/dataflow.rst
```diff
@@ -0,0 +1,685 @@
...
+    @df.region()
+    def top(A: float32[M, K], B: float32[K, N], C: float32[M, N]):
+        @df.kernel(mapping=[P0, P1], args=[A, B, C])
+        def gemm(local_A: float32[M, K], local_B: float32[K, N], local_C: float32[M, N]):
+            pi, pj = df.get_pid()
+            # Each kernel computes its tile
+            for i in range(pi * Mt, (pi + 1) * Mt):
+                for j in range(pj * Nt, (pj + 1) * Nt):
+                    for k in range(K):
+                        local_C[i, j] += local_A[i, k] * local_B[k, j]
```
Fangtangtang, line 375: I just realized that this tiling is probably better described using `layout` type.
The way it's written here almost looks like each program instance on a grid is computing a tile of the whole output tensor but writing the whole tensor back. feels like it has write conflicts.

but we do have test cases written like this in tests/dataflow/test_tiled_gemm.py

chhzh123: Good catch! Those tensors are actually not duplicated but shared across different grid points. Maybe a better way to specify this is to use global variables instead of creating local variables:

```python
    @df.region()
    def top(A: float32[M, K], B: float32[K, N], C: float32[M, N]):
        @df.kernel(mapping=[P0, P1])
        def gemm():
            pi, pj = df.get_pid()
            # Each kernel computes its tile
            for i in range(pi * Mt, (pi + 1) * Mt):
                for j in range(pj * Nt, (pj + 1) * Nt):
                    for k in range(K):
                        C[i, j] += A[i, k] * B[k, j]
```

wdyt?

Fangtangtang: yes! looks much clearer! 

One small point: if a tensor like `A` is passed via `args=[A]`, is it still allowed to access the region-level `A` inside the kernel?

chhzh123: No, it should be shadowed. Passing `args=[A]` and calling `A` inside is forbidden.

chhzh123: I'll update this interface in the next PR.

Fangtangtang: ok

chhzh123:
@Fangtangtang You need explicit approval for this PR if everything looks good

Fangtangtang (approved):

PR #554, opened 2026-02-26
fix: strip MLIR %alloc names from VHLS csim output and handle ap_int in nanobind wrapper

sunwookim028:
## Problem

When running `test_three_level_systolic_csim`, the Vitis HLS C++ emitter generates `kernel.cpp` containing raw MLIR SSA names like `%alloc[1]` — invalid C++ identifiers. This causes `g++` to fail with:

```
error: expected unqualified-id before '%' token
  int16_t %alloc[1];
```

A secondary issue: \[parse_cpp_function()\](cci:1://file:///home/sk3463/main/projects/allo/allo/backend/ip.py:30:0-112:17) in `ip.py` used `\w+` to match C++ types, which stops at `<`, so `ap_int<8>` was misread (only `8` was captured as the type name), producing invalid generated nanobind wrapper code.

## Fix

**`vitis.py`**: Add `re.sub(r'%(\w)', r'\1', hls_code)` at the start of \[postprocess_hls_code()\](cci:1://file:///home/sk3463/main/projects/allo/allo/backend/vitis.py:377:0-429:18). The MLIR `%` sigil always precedes a word character; the C++ modulo operator does not. Safe substitution.

**`ip.py`**: Broaden type-matching regex to handle `ap_int<N>`/`ap_uint<N>` types. Add \[resolve_nb_type()\](cci:1://file:///home/sk3463/main/projects/allo/allo/backend/ip.py:16:0-27:19) to map HLS types to `int{N}_t` for the nanobind interface, with `reinterpret_cast` back to the HLS type when calling the kernel.

## Testing

`tests/test_systolic_array.py::test_three_level_systolic_csim` now passes. This test only runs when Vitis HLS is available, which is why it was not caught by upstream CI.


chhzh123:
Can you add test cases for this?

sunwookim028:
Black formatting fixed. Added unit tests in `tests/test_backend_utils.py` covering `%alloc` stripping (including modulo-operator preservation) and `ap_int<N>`/`ap_uint<N>` → stdint type mapping. 9 tests, all passing.

sunwookim028:
Rebased onto current `main` (no longer behind). Test coverage in `tests/test_backend_utils.py` now at 12 cases covering `%alloc` stripping, modulo-operator preservation, and `ap_int<N>`/`ap_uint<N>` → stdint resolution. Ready for another look @chhzh123.

sunwookim028:
Hi @chhzh123! Just a gentle ping — it's been about 9 days since the last update. CI is green and all requested tests are in place (12 cases in `tests/test_backend_utils.py` covering `%alloc` stripping and `ap_int<N>`/`ap_uint<N>` resolution). Ready for review whenever you have a moment, thanks!

sunwookim028:
Hi @Fangtangtang, I have rebased this branch onto the current main and CI is green. Would you mind taking a look when you have a moment? Thanks in advance for the review.

Fangtangtang:
> Hi @Fangtangtang, I have rebased this branch onto the current main and CI is green. Would you mind taking a look when you have a moment? Thanks in advance for the review.

I noticed the unit test CI did not trigger due to a runner configuration issue. I have just fixed it (in #594). Could you please merge the latest main into your branch and rerun the CI? Thanks!

allo/backend/vitis.py
Fangtangtang, line 1: > the Vitis HLS C++ emitter generates `kernel.cpp` containing raw MLIR SSA names like `%alloc[1]` 

I think it would be better to fix this inside the HLS C++ emitter instead of relying on a string replacement in the postprocess step.
Could you help locate the bug in the `emit*HLS` files under the `mlir` directory? My guess is that there may be a bug in renaming steps.

tests/utils/test_backend_utils.py
Fangtangtang, line 1: Two suggestions for the unit tests:
1. Looks like the tests in this file are mainly for individual functions. We currently don't have a dedicated place for this kind of unit test under `tests/`, what about add a new folder (e.g., `tests/utils/`) to organize these tests? This may help keep the test structure cleaner.
2. I guess the issues you want to fix in this PR was discovered through using the IP module. Could we add a functional regression test for the actual feature being fixed (e.g., add an IP module example with the new datatype support under `tests/ip_integration`)?

Fangtangtang (changes requested):
Thanks for your work on this! Sorry for the delayed review. I left a few comments and suggestions below.

Also, could you also add some documentation (update `docs/source/dive/ip.rst`) for the supported data types in the IP module?

sunwookim028:
Thanks for the review! All three points addressed.

**Root cause fixed at the source, postprocessing removed.** You were right that this is a renaming bug, but it turned out to live one step before the emitter: `analyze_use_def` in `allo/passes.py` derives a fallback name from the textual SSA identifier (e.g. `%alloc`) and persists it into the op's `name` attribute without stripping the sigil; the HLS emitters then use that attribute verbatim. The fix strips the sigil where the name is derived, and the string-replacement step in `vitis.py` is removed. Verified on the `test_three_level_systolic_csim` schedule: the generated HLS code previously contained 80 `%`-prefixed identifiers (including paths the old regex never reached); it now contains none.

**Tests.** Moved the function-level unit tests to `tests/utils/` and replaced the `%`-stripping cases with an emission-level regression test (`test_no_ssa_sigil_in_emitted_hls_code`). Added a functional regression test for the new datatype support: `tests/ip_integration/vadd_ap_int.cpp` with `test_ap_int`, gated on Vitis HLS availability like `test_systolic_stream`. Verified locally with Vitis HLS 2023.2.

**Docs.** Added a "Supported data types" section to `docs/source/dive/ip.rst` covering the standard C types and `ap_int<N>`/`ap_uint<N>` (mapped to the matching fixed-width integer at the Python boundary; requires `link_hls=True`).

docs/source/dive/ip.rst
```diff
@@ -73,3 +73,29 @@ Similar to other Allo kernels, we can also change the build target to invoke the
 
     mod = s.build(target="vitis_hls", mode="csyn", project="vadd.prj")
     mod()
+
+Supported data types
+====================
+
+The element types allowed in the C++ IP signature are the standard C arithmetic types (``int``, ``float``, ``double``, and the fixed-width ``stdint`` types such as ``int8_t``/``uint16_t``), as well as the arbitrary-precision HLS integer types ``ap_int<N>`` and ``ap_uint<N>``. Because ``ap_int``/``ap_uint`` are provided by the Vitis HLS headers, an IP module that uses them must be created with ``link_hls=True``.
```
Fangtangtang, line 80: > provided by the Vitis HLS headers

shall we use "Vitis HLS libraries" instead to keep the terminology consistent with the documentation?

Also @chhzh123, I wonder if the naming `link_hls` should be more specific. Would `link_vitis_hls` be more accurate?

Fangtangtang (commented):
@chhzh123 Could you take a look at the `ap_int`/`ap_uint` support part (in `allo/backend/ip.py` and `docs/source/dive/ip.rst`) of this PR? I'm not very familiar with the HLS backend.

docs/source/dive/ip.rst
chhzh123: I think this should support different HLS backends, but currently, only Vitis HLS is supported, so for other backends, it should error out, and this could be another PR.

tests/utils/test_backend_utils.py
```diff
@@ -0,0 +1,166 @@
...
+
+    s = allo.customize(gemm)
+    buf_A = s.buffer_at(s.A, "i")
+    buf_B = s.buffer_at(s.B, "j")
+    pe = s.unfold("PE", [0, 1])
+    s.partition(s.C, dim=0)
+    s.partition(s.A, dim=1)
+    s.partition(s.B, dim=2)
+    s.to(buf_A, pe, axis=1, depth=M + 1)
+    s.to(buf_B, pe, axis=0, depth=N + 1)
```
chhzh123, line 52: Try to simplify this test case? Do we need so many primitives to reproduce the issue?

chhzh123 (approved):
The ap_int part looks good, just a minor issue about the test case

sunwookim028:
@Fangtangtang Took a while, but here is my fix: rebased on current main, clarified the Vitis HLS docs, and verified the `%alloc` regression. The retained test is the smallest schedule I found that reaches VHLS emission.

Fangtangtang (approved):
LGTM, thanks!

Issue #510, opened 2026-01-02
[Feature][Dataflow] Enrich Allo MLIR semantics by encoding dataflow metadata into IR


Fangtangtang:
### Motivation and Expected Behavior

The current dataflow representation in Allo IR includes the MLIR portion as well as some auxiliary data structures. 
Dataflow related information not encoded in Allo MLIR are:
- [x] Layout (shard, replicate): currently maintained via [DTensor class](https://github.com/cornell-zhang/allo/blob/2ea6b6b7a21d777ed31a5f23cb9afa1967de0d8a/allo/memory.py#L326)
- [x] Stream: currently maintained via maps and [Stream class](https://github.com/cornell-zhang/allo/blob/2ea6b6b7a21d777ed31a5f23cb9afa1967de0d8a/allo/backend/aie/utils.py#L105)
- [x] device mesh: currently maintained via maps

We would like to encode the above information directly into Allo MLIR

### Related resources

- https://github.com/openxla/shardy. 
- https://openxla.org/shardy/mpmd/mpmd_dialect
- https://mlir.llvm.org/docs/Dialects/Shard/

Fangtangtang:
[sdy.manual_computation](https://github.com/openxla/shardy/blob/main/docs/compiler_api.md#manual-computation) seems to be a good fit for expressing device mesh and layout (shard / replicate).

---

Current allo mlir for [vector addition on 4 tiles (compute units)](https://github.com/cornell-zhang/allo/blob/2ea6b6b7a21d777ed31a5f23cb9afa1967de0d8a/tests/dataflow/aie/test_vector.py#L214) is:
```mlir
module {
  func.func @core_0(%arg0: memref<256xi32>, %arg1: memref<256xi32>, %arg2: memref<256xi32>) attributes {df.kernel, itypes = "sss", otypes = "", stypes = "___", tag = "core_()"} {
    %alloc = memref.alloc() : memref<256xi32>
    linalg.add {op_name = "add_0"} ins(%arg0, %arg1 : memref<256xi32>, memref<256xi32>) outs(%alloc : memref<256xi32>)
    memref.copy %alloc, %arg2 {to = "local_C"} : memref<256xi32> to memref<256xi32>
    return
  }
  func.func @top(%arg0: memref<1024xi32>, %arg1: memref<1024xi32>, %arg2: memref<1024xi32>) attributes {dataflow, itypes = "sss"} {
    return
  }
}
```

We may rewrite the top fuction as
```mlir
[...]

PR #555, opened 2026-02-28
[MLIR][Dataflow] Add Allo operations for SPMW and setup support

Fangtangtang:
## Description ##
This PR adds new Allo operations for SPMW and updates the setup scripts accordingly.
It also includes several minor refactorings

**This changes are prerequisites for #510.**
#### New operations
1. `StreamGlobalOp`: Create a global stream or global stream array
2. `GlobalStreamGetOp`: Get an object from a global stream
3. `GlobalStreamPutOp`: "Put an object to a global stream
4. `GridMapOp`: create work grid

#### New passes
Added some useful passes fr lowering

1. `unroll`: unroll `meta_for` loops
2. `lower_memcopy_ops`: lower `memref.copy` to elementwise load-store. This is required for HLS backend.

### Minor Refactoring
- Renamed `stateful` to `Stateful` for naming consistency.
- Switched to the `register_dialect` mechanism for dialect registration.
- Added `externals/mlir-python-extra` as the MLIR Python binding library.
- Fixed issues in `pyproject.toml`.

## Checklist ##

- [ ] Code is well-documented

### DIFF
--- a/mlir/lib/Bindings/AlloModule.cpp
+++ b/mlir/lib/Bindings/AlloModule.cpp
@@ -373,7 +373,7 @@ NB_MODULE(_allo, m) {
   allo_m.def("unify_kernels", &UnifyKernels);
 
   allo_m.def("copy_on_write_on_function", &copyOnWriteOnFunction);
-  allo_m.def("unroll_loop", &unrollLoop, nb::arg("loop"),
+  allo_m.def("explicit_unroll", &unrollLoop, nb::arg("loop"),
              nb::arg("factor") = 0);
 
   // Utility APIs


mlir/lib/Conversion/LowerMemCopyOps.cpp
```diff
@@ -0,0 +1,162 @@
...
+      for (int64_t val : srcShape)
+        srcSize *= val;
+
+      int64_t dstSize = 1;
+      for (int64_t val : dstShape)
+        dstSize *= val;
+
+      if (srcSize != dstSize) {
+        return false;
+      }
```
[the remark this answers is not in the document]
Fangtangtang: invalidating cases when `srcSize`/`dstSize` are not positive

mlir/lib/Conversion/LowerMemCopyOps.cpp
```diff
@@ -0,0 +1,162 @@
...
+                  nestedBuilder.create<arith::ConstantIndexOp>(nestedLoc, 0);
+
+              for (unsigned d = 0; d < dstShape.size(); ++d) {
+                Value mul = nestedBuilder.create<arith::MulIOp>(
+                    nestedLoc, ivs[d],
+                    nestedBuilder.create<arith::ConstantIndexOp>(
+                        nestedLoc, dst_strides[d]));
+                dstIdx =
+                    nestedBuilder.create<arith::AddIOp>(nestedLoc, dstIdx, mul);
+              }
```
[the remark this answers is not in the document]
Fangtangtang: It's working fine

allo/ir/utils.py
```diff
@@ -104,7 +131,13 @@ def _adjust_line_numbers(node, offset):
             child.end_lineno += offset
 
 
-def parse_ast(src, starting_line_no=1, verbose=False):
+def parse_ast(src, verbose=False):
+    if isinstance(src, str):
+        starting_line_no = 1
+    else:
+        src, starting_line_no = inspect.getsourcelines(src)
+        src = [textwrap.fill(line, tabsize=4, width=9999) for line in src]
```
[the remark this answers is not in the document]
Fangtangtang: > `textwrap.fill()` in the Python standard library does not accept a `tabsize` keyword argument

It does accept `tabsize`. This is also the approach used in our existing codebase.

pyproject.toml
```diff
@@ -5,6 +5,8 @@
 requires = [
     "setuptools>=61.0",
     "wheel",
+    "cmake>=3.20",
+    "ninja",
```
Fangtangtang, line 9: Added this since dependencies may not always be found in the temporary build environment (similar to `nanobind`).

allo/ir/utils.py
```diff
@@ -27,19 +28,33 @@
 from .symbol_resolver import ASTResolver
 
 
-def _get_global_vars(_func):
+def _get_global_vars(_func, skip=None, stop=None):
```
chhzh123: Add comments: what does `skip` and `stop` mean

allo/ir/utils.py
```diff
@@ -52,13 +67,25 @@ def _get_global_vars(_func):
 
 
 def get_global_vars(func):
```
chhzh123, line 79: Any reasons why we want to change the `_get_global_vars` function?

externals/mlir-python-extra
chhzh123, line 1: This is for?

mlir/include/allo/Dialect/AlloOps.td
```diff
@@ -1048,6 +1048,56 @@ def StreamPutOp : Allo_Op<"stream_put"> {
   }];
 }
 
+def StreamGlobalOp : Allo_Op<"stream_global", [Symbol]> {
```
chhzh123, line 1051: Do we need to separate global streams and local streams? In our current definition, the streams are all global?

mlir/include/allo/Dialect/AlloOps.td
```diff
@@ -1096,4 +1146,35 @@ def TransformLayoutOp : Allo_Op<"transform_layout"> {
...
+//===----------------------------------------------------------------------===//
+// SPMW operations
+//===----------------------------------------------------------------------===//
+def GridMapOp : Allo_Op<"grid_map",
+  [RecursiveMemoryEffects, SingleBlockImplicitTerminator<"YieldOp">]> {
+  let summary = "Grid map operation";
+  let description = [{
+    The grid_map operation distributes a computation over a logical grid.
+    It takes a list of memrefs with static shape (`tensors`) as input arguments.
+    It shuns isolation from above and contains a single block whose arguments are the sharded memrefs.
```
chhzh123: What do you mean by `shuns isolation from above`? Maybe rewrite this sentence

mlir/lib/Transforms/LoopTransformations.cpp
```diff
@@ -3718,6 +3718,18 @@ bool applyLoopTransformationOnSingleFunction(
...
+  func->walk<WalkOrder::PostOrder>([&](AffineForOp forOp) {
+    auto attr = forOp->getAttrOfType<StringAttr>("loop_type");
+    if (attr && attr.getValue() == "unroll") {
+      if (failed(loopUnrollFull(forOp))) {
+        forOp.emitError("failed to fully unroll the loop");
+      }
+    }
+  });
+  return true;
+}
```
chhzh123: We have `runUnrolling`.  Why do we need an additional `applyUnroll` function?

chhzh123: I suppose this should be applied using schedule primitives

setup.py
```diff
@@ -36,6 +37,25 @@ def build_extension(self, ext):
...
+                [
+                    "git",
+                    "submodule",
+                    "update",
+                    "--init",
+                    "--recursive",
+                    "externals/mlir-python-extra",
+                ],
+                cwd=Path(ext.sourcedir).parent,
+            )
```
chhzh123: It's indeed weird to run `git pull` in the setup script. Does it have a pre-built Python library that we can leverage? Otherwise, we should write in the doc to let the users pull the submodules first

mlir/include/allo/Dialect/AlloOps.td
Fangtangtang: > In our current definition, the streams are all global?

The key difference is between Symbols and SSA values.
our current StreamConstructOp returns an SSA value. Even if the stream is defined directly under the top module, it cannot be directly used by operations inside a function also under the same module.

Fangtangtang: > Do we need to separate global streams and local streams?

I don't want to. 😓
However, with the current stream ops it's hard to model a stream shared across functions, so I want new opertions. At the same time, all the dataflow backend lowering passes are tightly coupled with current stream operations. replacing these operations would require rewriting all those passes first.

externals/mlir-python-extra
Fangtangtang: the MLIR Python binding library

mlir/lib/Transforms/LoopTransformations.cpp
Fangtangtang: this is used to unroll all the `meta_for` loops. I'll try if I can directly use the primitives

mlir/include/allo/Dialect/AlloOps.td
chhzh123: I see, but in the end, we don't want to keep two different copies of the same operations. I guess MLIR should have some ways to resolve this

externals/mlir-python-extra
chhzh123: I meant what specific facility in this Python binding library is needed?

allo/ir/utils.py
Fangtangtang: Mainly to support multi-level imports. For example, A imports B, and B further imports C, so `_get_global_vars` needs to handle global variables coming from imported modules as well.

externals/mlir-python-extra
Fangtangtang: These are mainly some wrappers. They are not used by the current code, but the new builder will rely on some of these wrappers.

chhzh123: Then maybe leave for the next PR?

Fangtangtang: ok

setup.py
Fangtangtang: also leave for the next PR

mlir/lib/Transforms/LoopTransformations.cpp
Fangtangtang: > We have `runUnrolling`.

This is used to unroll one loop in a function. I can move the 'traverse and apply rununrolling' to python.

Also wondering what would happen if nested loops both need to be unrolled, and we apply unrolling to the outer loop first? feels like this would cause issues. is this something that should somehow be avoided by the ordering of schedule primitives?

mlir/include/allo/Dialect/AlloOps.td
Fangtangtang: > it cannot be directly used by operations inside a function also under the same module.

forgot to mention, this is because `FuncOp` has `IsolatedFromAbove` trait

mlir/lib/Transforms/LoopTransformations.cpp
chhzh123: What kind of issues? The final transformation result should not be affected by which loop is unrolled first

Fangtangtang: after unrolling the outer loop, there will be multiple copies of the inner loop

Fangtangtang: if we use some handle for the loop to be unrolled, will the handle be invalid? 
if we use loop's name attributes to find the loop to be unrolled, we may need to give all the loop copies the same name and  'match' all the loops with this name.

Fangtangtang: oh, this is fine as a scheduling primitive. The loop is not actually unrolled in the IR, it just adds an `unroll` attribute that will be consumed by the backend, for example lowered to a `pragma` in the HLS backends.

chhzh123: so do you need the loop to be explicitly unrolled or not? If you need explicit unroll, then maybe use the `unfold` primitive.

In theory, after applying the unroll primitive, it should return all the handles of the unrolled loops, so they can be further chained with other primitives

Fangtangtang: This primitive may not be suitable for unrolling `meta_for` then. `meta_for` should be unrolled directly in the IR.

chhzh123: Use [`unfold`](https://github.com/cornell-zhang/allo/blob/main/allo/customize.py#L1018)?

Fangtangtang: What about write a new method in `LoopTransformations.cpp` using [`loopUnrollFull`](https://github.com/llvm/llvm-project/blob/aa89092c0219709775fdc0957095aeb919afb1c2/mlir/lib/Dialect/Affine/Utils/LoopUtils.cpp#L871) or [`loopUnrollUpToFactor`](https://github.com/llvm/llvm-project/blob/aa89092c0219709775fdc0957095aeb919afb1c2/mlir/lib/Dialect/Affine/Utils/LoopUtils.cpp#L886), expose python binding interface, then use it for both [unfold](https://github.com/cornell-zhang/allo/blob/main/allo/customize.py#L1018) and `meta_for` unrolling.

reusing MLIR's unrolling utilities to unroll a single loop would be cleaner. I'd prefer not to handle `meta_for` unrolling by naming loops and then matching by name.

Fangtangtang: > use it for both [unfold](https://github.com/cornell-zhang/allo/blob/main/allo/customize.py#L1018) and meta_for unrolling.

sorry, I realized this won't work.😅 `unfold` is doing more than loop unrolling, it also duplicates the callee of call operations inside the loop.

chhzh123: ok, anyway, you can introduce explicit loop unrolling, but just make it consistent with other passes in `LoopTransformations.cpp`. Or maybe create a separate pass dedicated to SPMW, as Kai at some point will remove the current design of LoopTransformation

mlir/lib/Bindings/AlloModule.cpp
```diff
@@ -361,6 +373,8 @@ NB_MODULE(_allo, m) {
   allo_m.def("unify_kernels", &UnifyKernels);
 
   allo_m.def("copy_on_write_on_function", &copyOnWriteOnFunction);
+  allo_m.def("unroll_loop", &unrollLoop, nb::arg("loop"),
+             nb::arg("factor") = 0);
```
Fangtangtang: I want to expose loop unroll. The `factor` argument is optional, when not provided (default = 0), the loop will be fully unrolled.

```python
import allo._mlir.dialects.allo as allo_d
import allo._mlir.dialects.affine as affine_d

...
if isinstance(op1, affine_d.AffineForOp):
    allo_d.unroll_loop(op1) # fully unroll
if isinstance(op2, affine_d.AffineForOp):
    allo_d.unroll_loop(op2, 0) # fully unroll    

if isinstance(op3, affine_d.AffineForOp):
    allo_d.unroll_loop(op3, 4) # unroll by 4
```

mlir/lib/Bindings/AlloModule.cpp
```diff
@@ -361,6 +373,8 @@ NB_MODULE(_allo, m) {
   allo_m.def("unify_kernels", &UnifyKernels);
 
   allo_m.def("copy_on_write_on_function", &copyOnWriteOnFunction);
+  allo_m.def("unroll_loop", &unrollLoop, nb::arg("loop"),
```
chhzh123: Let's rename it as `explicit_unroll`

chhzh123 (approved):
A small naming issue. Otherwise LGTM

mlir/include/allo/Dialect/AlloOps.td
```diff
@@ -1096,4 +1146,35 @@ def TransformLayoutOp : Allo_Op<"transform_layout"> {
...
+  let description = [{
+    The grid_map operation distributes a computation over a logical grid.
+    It takes a list of memrefs with static shape (`tensors`) as input arguments.
+    It contains a single block whose arguments are the sharded memrefs. The operations in the block can access variables in parent regions.
+  }];
+
+  let arguments = (ins
+    Variadic<AnyStaticShapeMemRef>:$tensors,
+    ArrayAttr:$sharding,
+    DenseI64ArrayAttr:$grid
```
Fangtangtang, line 1164: alternatives:
- [Affine Map](https://mlir.llvm.org/docs/Dialects/Affine/#affine-maps) for `sharding`
- [Integer Set](https://mlir.llvm.org/docs/Dialects/Affine/#integer-sets) for `grid`

Issue #556, opened 2026-03-03
[Bug][HLS] Compile error `<func_name> was not declared in this scope` in HLS backend


Fangtangtang:
### Describe the bug

In the HLS backend, we encountered a compile-time error for some test cases: `<func_name> was not declared in this scope`

This is caused by a forward reference in the generated HLS C++ kernel code.

Forward references are valid in MLIR (functions can be called before their definition appears in the module). However, this is not valid in cpp-style HLS kernels unless a proper function declaration is emitted before use.
Currently, due to sequential lowering, the generated kernel may contain a function call that appears before the function definition, leading to the compile error above.

### Reproduction

Run test case in `tests/dataflow/test_hierachical.py`

### Buggy output

```txt
RuntimeError: Failed to compile nanobind wrapper for pytop_1772560633021751462! In file included from allo/tests/dataflow/top.prj/_tmp/pytop_1772560633021751462.cpp:6:
allo/tests/dataflow/top.prj/kernel.cpp: In function ‘void wrapper_0(float (*)[32], float (*)[32], float (*)[32], float (*)[32])’:
allo/tests/dataflow/top.prj/kernel.cpp:26:3: error: ‘inner_2_2_0’ was not declared in this scope
   26 |   inner_2_2_0(v0, v1, v2);      // L3
      |   ^~~~~~~~~~~
allo/tests/dataflow/top.prj/kernel.cpp: In function ‘void wrapper_1(float (*)[32], float (*)[32], float (*)[32], float (*)[32])’:
allo/tests/dataflow/top.prj/kernel.cpp:126:3: error: ‘inner_4_4_1’ was not declared in this scope; did you mean ‘inner_2_2_0’?
  126 |   inner_4_4_1(v55, v56, v58);   // L74
      |   ^~~~~~~~~~~
      |   inner_2_2_0
```

### Expected behavior

should compile successfully

#### 1. Quick Fix: Reorder Functions by Call Graph
Introduce a pass that: Builds the call graph of the MLIR module and reorders `func.func` operations in topological order.

#### 2. Proper Fix: Emit Function Declarations Before Definitions
Modify all HLS codegen passes (all `EmitHLS.cpp` files) to: emit function declarations for all functions before full function definitions.

chhzh123:
Why is the first option not a "proper" fix? I think it'd be better to separate transformations from the codegen, which means we should reorder the functions before codegen

Fangtangtang:
[...]

PR #557, opened 2026-03-03
[Fix] several issues related to `tests/dataflow/test_hierachical.py`

Fangtangtang:
## Description ##
This PR fixes issues related to `tests/dataflow/test_hierachical.py`. 

### Problems ###
1. Incorrect path resolution in `IPModule`: The project path passed into `IPModule` was not properly resolved, which caused incorrect file lookup when the HLS backend was set to `csim` mode. 

2. #556: Function forward references caused compilation failures in HLS backends.

3. output argument number is not correctly analyzed. Some mode would fail.

### Proposed Solutions ###
1. The `project` argument is now internally converted to an absolute path for correct file resolution.
2. Partially address forward reference issue
   * For VivadoHLS, function declaration emission has been added so that all functions are declared before use.
   * For other backends, a function reordering pass has been introduced in `passes.py`, which reorders functions based on the call graph to avoid forward references. 

3. reuse `analyze_arg_load_store` for argument input / output type analysis.

## Checklist ##

- [ ] Code is well-documented


Fangtangtang:
> 2. Partially address forward reference issue
>    * For VivadoHLS, function declaration emission has been added so that all functions are declared before use.
>    * For other backends, a function reordering pass has been introduced in `passes.py`, which reorders functions based on the call graph to avoid forward references. 

we should consider adding declaration emission support to other HLS backends

Fangtangtang:
simplified a bit. a declaration is only emitted when the function is defined after its earliest caller. This makes the generated hls code cleaner.

tests/ip_integration/test_external.py
```diff
@@ -5,13 +5,16 @@
...
+from pathlib import Path
 import allo
 from allo.ir.types import int32, float32
 import allo.backend.hls as hls
 
 
 def test_nanobind():
-    mod = allo.IPModule(top="gemm", impl="gemm.cpp", link_hls=False)
+    mod = allo.IPModule(
+        top="gemm", impl=Path(__file__).resolve().parent / "gemm.cpp", link_hls=False
```
chhzh123, line 16: So the original relative path does not work? I don't want the users to provide absolute paths all the time

Fangtangtang: A relative path would work if it correctly points from the test working directory to the implementation path. 
In this case, the files are organized
```txt
ip_integration
├── test_external.py
└── gemm.cpp
```
`impl="gemm.cpp"` works if we are running the test under `ip_integration` directory.

allo/passes.py
```diff
@@ -987,3 +987,62 @@ def pipe_loop_innermost(forop, ii, rewind):
...
                         pipe_loop_innermost(op_, ii, rewind)
+
+
+def reorder_funcs(module):
+    """
+    Reorder FuncOps in a module so that callees appear before callers.
+
+    [NOTE]: this is a patch for HLS backend to avoid forward reference errors in generated C++ code.
+    This pass performs a topological sort on the call graph and moves FuncOps so that every function is defined before it is called.
+    """
```
chhzh123: I thought we agreed that only the declaration needs to be printed out at the beginning, instead of reordering the functions?

chhzh123 (commented):
Also let me know if you finish the changes

allo/passes.py
Fangtangtang: I only added declaration generation for emitVivadoHls. So this is not used if we use `allo_d.emit_vhls` for codegen. 
I left the pass here in case other hls backend have the same issue

allo/backend/hls.py
```diff
@@ -251,6 +251,8 @@ def __init__(
         buf = io.StringIO()
         success = True
         match platform:
+            # [NOTE]: If the HLS backend reports "<func_name> was not declared in this scope", it is likely caused by a forward reference.
+            #           MLIR allows calling functions before their definition, but C++ HLS kernels require a prior declaration.
```
Fangtangtang, line 255: The reorder pass is removed. I added comment here instead.

Fangtangtang:
finished

chhzh123 (approved):
LGTM, thanks

Issue #561, opened 2026-03-06
[Bug] Hierarchical regions with inter-kernel streams deadlock in simulator and have HLS codegen issues


sunwookim028:
### Description

When using hierarchical dataflow regions (a `top` region calling an `inner` region via `inner(...)` inside a wrapper kernel), both the **simulator** backend and the **Vitis HLS** backend fail when the inner region contains kernels that communicate via streams with circular dependencies (e.g., producer sends data, then waits for consumer's processed result).

### Reproducer

```python
import allo
from allo.ir.types import float32, Stream
import allo.dataflow as df
import numpy as np

@df.region()
def inner(A: float32[4], B: float32[4]):
    fwd: Stream[float32, 4][1]
    bwd: Stream[float32, 4][1]

    @df.kernel(mapping=[1], args=[A])
    def producer(local_A: float32[4]):
        fwd[0].put(local_A[0])
        fwd[0].put(local_A[1])
        r0: float32 = bwd[0].get()  # Wait for consumer result
        r1: float32 = bwd[0].get()
        local_A[2] = r0
        local_A[3] = r1

    @df.kernel(mapping=[1], args=[B])
    def consumer(local_B: float32[4]):
        v0: float32 = fwd[0].get()
        v1: float32 = fwd[0].get()
        local_B[0] = v0 + 1.0
        local_B[1] = v1 + 1.0
        bwd[0].put(v0 + 1.0)
        bwd[0].put(v1 + 1.0)

@df.region()
def top(A: float32[4], B: float32[4]):
    @df.kernel(mapping=[1], args=[A, B])
    def wrapper(local_A: float32[4], local_B: float32[4]):
        inner(local_A, local_B)

# WORKS: flat build, simulator runs kernels in parallel
sim = df.build(inner, target="simulator")
A = np.array([10.0, 20.0, 0.0, 0.0], dtype=np.float32)
B = np.zeros(4, dtype=np.float32)
sim(A, B)  # Passes

# DEADLOCKS: hierarchical build, simulator serializes inner kernels
sim2 = df.build(top, target="simulator")
sim2(A, B)  # Hangs forever

# HLS CODEGEN BUG: wrapper_0 calls inner__0 before it's defined
hls_mod = df.build(top, target="vitis_hls", mode="csim", project="test.prj")
hls_mod(A, B)  # Compilation error: 'inner__0' was not declared
```

#### 1. Simulator Backend: Deadlock

[...]

PR #562, opened 2026-03-06
fix: resolve simulator deadlock for hierarchical dataflow regions

sunwookim028:
## Problem
`df.build(top, target="simulator")` deadlocks when a `top` region wraps an inner region whose kernels communicate via streams (#561).
**Root cause**: `build_dataflow_simulator` only injected `omp.parallel`/`omp.sections` around the top-level function's `func.call` ops. Inner region kernel calls remained sequential.
## Fix
- Extract OMP injection into \[_inject_omp_parallel_sections()\](cci:1://file:///home/sk3463/main/projects/allo/allo/backend/simulator.py:780:0-803:45) helper
- Add `all_pe_calls_by_func` accumulator to `_process_function_streams` — collects PE calls per-function during recursive traversal
- `build_dataflow_simulator` now injects OMP sections into **every** function with PE kernel calls
## Testing
- `test_1x1_compute_tile` ✅ — flat build, no regression
- `test_1x1_hierarchical` ✅ — hierarchical build now works (previously deadlocked)
- `test_hierachical` ✅ — simulator portion passes

Addresses a part of #561

### DIFF
--- a/allo/backend/simulator.py
+++ b/allo/backend/simulator.py
@@ -79,6 +79,7 @@ def _process_function_streams(
     module: Module,
     func: func_d.FuncOp,
     processed_funcs: set,
+    all_pe_calls_by_func: dict = None,
 ):
     """
     Process streams and PE calls within a single function.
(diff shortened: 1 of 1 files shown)


chhzh123:
Can you fix the CI? and it'd be good to have additional test cases

PR #563, opened 2026-03-06
Fix HLS codegen for hierarchical dataflow regions (#561)

### LINKED ISSUE #561: [Bug] Hierarchical regions with inter-kernel streams deadlock in simulator and have HLS codegen issues

### LINKED ISSUE #565: [Bug] Missing global_op_cache on ASTContext.copy()
### Description
When `df.region()` compiles nested kernels that use `@stateful` streams, it relies on mapping global names to MLIR operations. During compilation, the context for nested blocks is duplicated via `ASTContext.copy()`. However, `copy()` previously failed to selectively copy the `global_op_cache` dictionary, leading to `AttributeError` when the compiler attempted to locate `global_op_cache` for stateful variable generation.

### Reproducing Steps
1. Create a `df.region()`
2. Define a nested kernel inside using `@df.kernel(...)`
3. Declare an internal stream inside the nested kernel using `val: type @ stateful = 0`
4. Compile using `df.build(region_name)`

### Proposed Fix
Explicitly copy the `global_op_cache` attribute if it exists when duplicating the context in `allo/ir/visitor.py`.

### Files affected
- `allo/ir/visitor.py`

sunwookim028:
This PR resolves HLS compilation errors ('undeclared identifier') by ensuring all functions are forward-declared in the generated C++ code. It also ensures nested dataflow regions receive the correct HLS pragmas. Closes #561.

Closes #565.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -21,3 +21,24 @@
 # Don'ts
 - Do not modify repository structure without approval
 - Do not install system packages without explicit user confirmation
+
+# Troubleshooting & Known Issues
+- **"Fail to resolve the expression as symbolic expression" in Dataflow**: When using stream arrays (e.g., `gemm_in_A[m]`), the index `m` must be a compile-time constant (like `df.get_pid()`) or statically unrollable. Using a dynamic runtime loop index or variable will cause this type-inference failure. Manually unroll the loops or use literal constants where possible.
+- **"AttributeError: 'ASTContext' object has no attribute 'global_op_cache'"**: This occurs when compiling stateful variables in nested kernels inside `df.region()`. Fixed upstream; ensure `ASTContext.copy()` properly preserves `self.global_op_cache`. Documented in `ALLO_CHANGE.md`.
+- **"func.call op operand type mismatch: expected operand type 'i32', but provided 'memref<i32>'"**: When passing scalars into a `df.region()` top-level function, MLIR's `_build_top` lowers the arguments to memrefs, causing a crash when passed to inner scalar-expecting implementations. **Workaround**: Always type scalar arguments in `df.region()` and `df.kernel()` as 1-element arrays, e.g., `size: int32[1]` instead of `int32`, and access via `size[0]`.
+- **Duplicate annotated variable names cause silent hangs**: In Allo, `x: int32 = expr` is an annotated assignment that creates a new buffer. If the same variable name is annotated twice in the same kernel (e.g., `addr1: int32 = ...` in two separate `if` branches), Allo may silently hang during compilation or simulation. **Fix**: Use unique names for each branch (e.g., `g_addr1` / `v_addr1`).
+- **Dataflow deadlock from conditional stream usage**: In `df.region()`, all inner kernels execute exactly once per invocation. If one kernel conditionally `.put()`s to a stream but another kernel unconditionally `.get()`s from the same stream, the reader blocks forever when no data is sent. **Fix**: Use explicit enable streams (`en: Stream[int32, depth]`) that are always `.put()` by the producer and always `.get()` by the consumer. Guard the actual data stream `.get()` behind `if en == 1:`.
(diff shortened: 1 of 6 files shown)


chhzh123:
Can you test Shihan's PR https://github.com/cornell-zhang/allo/pull/557 first?

sunwookim028:
Thanks, I did.

chhzh123:
CI needs to be fixed

sunwookim028:
Closing in favor of a minimal focused branch. The new PR contains only the three targeted fixes for #561 and #565, cherry-picked onto current main.

PR #568, opened 2026-03-15
[Python] Introduce custom lightweight Python bindings with semantic builder APIs

kkkaishao:
This is the second PR of the PR stack #567, #568, #569

## Description ##
This PR introduces custom Python bindings for Allo and shifts Python-side IR construction from the traditional upstream MLIR binding style to a more semantic, C++-API-inspired programming model.

The new bindings expose core IR objects and commonly used dialect operations through a lightweight Allo-specific API surface. Instead of manually juggling raw upstream operation classes, insertion points, locations, and dialect-specific boilerplate, users can now construct IR through semantic wrappers and `AlloOpBuilder`, which makes IR building easier to read, easier to maintain, and easier to distribute.

### Motivation ###
Before this change, Python IR construction in Allo primarily followed the upstream MLIR Python binding model. That model is powerful, but it is also relatively low-level:
- operation construction is close to raw MLIR op signatures
- insertion point management is more manual and scattered
- location handling is repetitive
- Python-side ergonomics are not tailored to Allo users
- packaging is heavier because the user model inherits more of the upstream binding organization

This PR addresses these issues by introducing an Allo-native binding layer that is:
- more semantic
- easier to use correctly
- lighter to package and distribute
- better suited for IDE support and long-term API evolution

### Problems ###
This PR focuses on the following problems:

1. Python IR construction relied too much on the upstream MLIR binding style, which exposed low-level details directly to users and internal builder code.
2. Insertion point management was correct but verbose, requiring direct manipulation of block bodies and `InsertionPoint(...)` objects.
3. The API surface was not sufficiently semantic: many operations were accessed as generic upstream ops instead of typed wrappers with domain-relevant construction/access patterns.
4. Python packaging and distribution were heavier than necessary for Allo-specific use cases.
5. Developer experience was limited by weaker type hints and less discoverable APIs.

### Proposed Solutions ###
This PR introduces a custom Python binding stack for Allo with the following changes:

1. Provide Allo-specific Python bindings for core upstream MLIR IR concepts:
   - `Context`
   - `Operation`
   - `Block`
   - `Region`
   - types and attributes commonly used in Allo workflows

2. Introduce `AlloOpBuilder`, which:
   - stores and reuses context/location state
   - supports explicit but simple insertion point control
   - removes much of the repeated `ip=...` plumbing in Python IR construction
   - makes builder-style IR generation feel closer to the C++ API

3. Provide semantic wrappers for commonly used upstream dialect operations:
   - `arith`
   - `func`
   - `affine`
   - `scf`
   - `tensor`
   - `memref`
   - `linalg`
   - `transform`
   - and other core dialects added in this series

4. Add semantic construction helpers and typed accessors:
   - operations can be constructed from an `AlloOpBuilder`
   - results, bodies, blocks, and attributes are accessed through typed wrapper APIs
   - `isa` / `dyn_cast` / `cast` now work over wrapped operations and prefer the most specific wrapper when appropriate

5. Add lazy-loading for submodules:
   - improves import-time performance
   - avoids eagerly loading every dialect wrapper
   - keeps the Python API responsive and modular

6. Improve developer tooling:
   - add comprehensive `.pyi` stubs
   - improve IDE auto-completion
   - improve static type information during development

7. Make distribution lightweight:
   - bindings are packaged as a single `._liballo` shared library
   - the resulting deployment model is simpler and better suited for distribution

### User-Facing Changes ###
The main user-visible change is that Python-side IR building becomes more semantic and much less tied to the raw upstream MLIR binding style.

Old style characteristics:
- construct ops directly from upstream dialect bindings
- pass `ip=ctx.get_ip()` repeatedly
- manually create and move `InsertionPoint(...)`
- use lower-level op/result handling patterns throughout the code

New style characteristics:
- construct IR through `AlloOpBuilder`
- set insertion points once, then build naturally
- create typed ops directly from semantic wrappers
- use typed `get_result_at`, `get_body`, `add_entry_block`, `isa`, `dyn_cast`, and `cast`
- interact with operations through wrapper classes instead of generic raw op objects

### Why This Is Better ###
This change improves the Python frontend in several concrete ways:

1. More semantic APIs
   - Users write code against Allo concepts and typed wrappers instead of raw upstream binding details.
   - The code reads more like an intentional IR construction program and less like manual assembly of MLIR internals.

2. Simpler insertion point management
   - `AlloOpBuilder` centralizes insertion point and location handling.
   - This removes repeated low-level boilerplate and reduces mistakes around where new operations are inserted.

3. Better ergonomics for IR construction
   - The API is closer to the C++ builder model.
   - This makes it easier to write new frontend utilities, transformations, and tests.

4. Better typed behavior
   - `isa` / `dyn_cast` / `cast` over wrapped operations make traversal and rewriting cleaner.
   - Shared op names can still resolve to the more specific concrete wrapper.

5. Better import and development experience
   - Lazy loading reduces unnecessary module initialization.
   - `.pyi` files improve editor support and discoverability.

6. Better distribution story
   - A single shared library is easier to package and ship.
   - This is especially helpful for lightweight deployment and downstream reuse.

### Examples ###
Below is a representative example of the new API style from `tests/test_bindings_dyn_cast.py`.

New custom bindings style:
```python
from allo.bindings import arith, func, ir, linalg, memref, scf, tensor

ctx = ir.Context()
ctx.load_dialects()

builder = ir.AlloOpBuilder(ctx)
builder.set_unknown_loc()

mod = ir.ModuleOp(builder)
builder.set_insertion_point_to_start(mod.get_body())

func_type = ir.FunctionType.get([], [], ctx)
func_op = func.FuncOp(builder, "test_dyn_cast", func_type)
entry = func_op.add_entry_block()

i32 = ir.IntegerType.get(32, ctx)
tensor_type = ir.RankedTensorType.get([4], i32)
identity_map = ir.AffineMap.get_identity(1, ctx)
memref_type = ir.MemRefType.get([4], i32, identity_map)

builder.set_insertion_point_to_end(entry)
const_i32 = arith.ConstantIntOp(builder, i32, 7)
lb = arith.ConstantIndexOp(builder, 0)
ub = arith.ConstantIndexOp(builder, 4)
step = arith.ConstantIndexOp(builder, 1)

loop = scf.ForOp(
    builder,
    lb.get_result_at(0),
    ub.get_result_at(0),
    step.get_result_at(0),
    [],
)
builder.set_insertion_point_to_end(loop.get_body())
scf.YieldOp(builder, [])

builder.set_insertion_point_to_end(entry)
tensor_empty = tensor.EmptyOp(builder, tensor_type)
memref_alloc = memref.AllocOp(builder, memref_type)
memref_load = memref.LoadOp(
    builder, memref_alloc.get_result_at(0), [lb.get_result_at(0)]
)

func.ReturnOp(builder, [])

# isa/dyn_cast APIs
casted = ir.cast(op, func.FuncOp)
ir.isa(op, func.FuncOp)
assert ir.dyn_cast(op, scf.ForOp) is None
```

Expected behavior and benefits:
- the builder carries construction context naturally
- insertion point updates are explicit but compact
- operations are created through typed Allo wrappers
- operation inspection uses typed wrappers and casting utilities
- IR construction is easier to read and less error-prone

For comparison, the previous style used throughout the existing builder stack is much closer to raw upstream MLIR Python bindings. Typical patterns include:
```python
func_op = func_d.FuncOp(name=node.name, type=func_type, ip=ctx.get_ip())
alloc_op = memref_d.AllocOp(memref_type, [], [], ip=ctx.get_ip())
affine_d.AffineYieldOp([], ip=InsertionPoint(for_op.body))
scf_d.YieldOp([], ip=InsertionPoint(for_op.body))
```

Compared with the old style, the new bindings:
- remove repeated `ip=ctx.get_ip()` plumbing
- reduce direct manipulation of low-level insertion-point objects
- provide more semantic op wrappers and typed APIs
- make operation construction and traversal significantly clearer

### Packaging / Distribution ###
This PR also improves how the bindings are packaged and distributed.

The binding module is built as a single shared object:
- `allo/bindings/_liballo`

This has two practical benefits:
- simpler distribution and deployment
- lighter-weight packaging for users who only need the Allo Python binding layer

### Compatibility ###
This PR is additive from the perspective of introducing a new Python binding layer. It does not attempt to remove the existing frontend builder infrastructure in this series, but it establishes a cleaner and more maintainable API for future Python-side IR construction and tooling.

### Testing ###
This PR includes:
- binding implementation tests for typed casts and wrapper behavior
- regression coverage for operation wrapping across multiple dialects
- checks that `isa`, `dyn_cast`, and `cast` behave correctly for wrapped ops
- tests that specific wrappers are preferred when multiple wrappers share an op family

Representative coverage includes:
- `func.FuncOp`
- `scf.ForOp`
- `tensor.ExtractOp`
- `memref.LoadOp`
- `linalg.GenericOp`
- concrete-vs-generic constant op casting behavior

## Checklist ##

- [ ] Code is well-documented


Fangtangtang:
Thanks a lot for your contribution and detailed PR description! The motivation is very clear, and I agree that improving the Python-side developer experience is important.

But I do have some concerns about introducing a custom Python binding layer that replaces the upstream MLIR bindings.
**Maintainability**: MLIR evolves quickly, and we won't stick to a fixed version. If we maintain our own binding implementation for core IR concepts, seems that we have to continuously track and adapt to those upstream changes?
**Compatibility** with the broader MLIR ecosystem. Using the upstream bindings makes it easier to interoperate with other MLIR-based projects or out-of-tree dialects. If the new bindings cannot coexist with the upstream ones, it could make such integrations hard. Allo uses mlir-aie (an out-of-tree dialect) as a backend, this could be an issue. 
More generally, we prefer to **avoid reimplementing infrastructure that already exists upstream** unless there is a strong reason to do so. If certain parts of the upstream infrastructure are difficult to use, it might also be worth considering improving them upstream?

kkkaishao:
> Maintainability: MLIR evolves quickly, and we won't stick to a fixed version.

- Yes, and maintaining a full alternative to the upstream MLIR Python bindings is definitely a bad idea. What I think is reasonable is a much narrower scope: this layer should primarily serve **frontend IR construction, schedule authoring, and other Allo-specific authoring workflows (as a glue layer)**. 
- In practice, the MLIR APIs that evolve most quickly are usually not the core IR construction APIs themselves.
The parts that tend to change more aggressively are complex analyses, rewrites, pass infrastructure, and dialect-specific helper layers. By contrast, the core concepts we rely on here, such as operations, values, blocks, regions, types, attributes, insertion points, and locations, are quite stable now. So I think the maintenance cost is acceptable if we keep this layer disciplined and centered on IR construction rather than heavy analysis or heavy Python-side pass development.
- If this custom layer were intended to support complex Python-side pass writing in a broad sense, I would share the concern much more strongly.

> Compatibility with the broader MLIR ecosystem.

- I think this is the more important concern. For interoperability, my view is that we should not require the custom Python layer to model every external dialect directly. In fact, if compatibility is a priority, then for almost all out-of-tree dialects, the better long-term integration path is not “wrap their Python bindings too,” but rather “integrate through C++ lowering / conversion passes.”
- I looked at the current AIE backend under `allo/backend/aie`, and it currently relies quite heavily on upstream / mlir-aie Python bindings to:
  - inspect and rewrite IR
  - build AIE dialect operations
  - run AIE pass pipelines
  - generate target-specific IR structures such as tiles, FIFOs, runtime sequences, etc.
- That dependency is real, but I do not think it argues against the custom bindings themselves. Instead, I think it highlights a natural architectural boundary: target-specific lowering into external dialect ecosystems is better implemented as C++ conversion / lowering passes, not as Python-side IR construction logic.
- In other words, for compatibility with out-of-tree dialects, my preferred direction would be:
   - use the custom Python bindings for Allo frontend construction and scheduling
   - implement out-of-tree dialect integration primarily through C++ conversion pipelines
   - expose only a simple Python API on top of those pipelines
- I think this is a better compatibility story than trying to make the custom Python bindings coexist deeply with every external dialect’s own Python bindings. If the interoperability point is “MLIR module in, MLIR module out, lowering done by C++ passes,” then the custom bindings do not need to understand each external dialect in detail.

> More generally, we prefer to avoid reimplementing infrastructure that already exists upstream
- Agree with it. In this case, I think the strong reason is frontend ergonomics, not replacing upstream infrastructure wholesale. 
- I also agree that, where the pain points are genuinely general, it is worth considering upstream improvements. But the reality is, I haven't contributed to the upstream, and I'm not sure whether upstream (which is built for general needs) really needs these things😂. For example, building IR with `OpBuilder` and managing the insertion point internally like C++ does is natural🤔, but I didn't find any discussion about it upstream. If some utilities are worth moving them upstream, it will be great.

So overall, the conclusion is:  a custom Python binding layer is reasonable if its scope is kept narrow, and if we don't need to support complex pass development on the Python side (basic ones are ok to handle though I do not recommend), the maintainability and compatibility concerns become much more manageable.

The main decision point is whether we need to support pass development with Python. 🤔

My long-term model is:
- Python for frontend ergonomics (act as glue, orchestrate the working flow)
- C++ for target-specific lowering, ecosystem interoperability, complex analysis and rewriting
- MLIR IR / pass pipelines as the compatibility boundary

> If the new bindings cannot coexist with the upstream ones

The situation is, if the new bindings library is dynamically linked, then it may work with the existing bindings (but I didn't fully test that). It requires some parts of upstream libraries are loaded before (to provide some symbols), or you can say that it just cannot work independently. If the new bindings library is statically linked, it can work without existing bindings and cannot work with it. In this situation, it will be easy to distribute Allo as a Python library, since we don't need carry lots of compilation artifacts from the upstream, only with an all-in-one *.so library

Fangtangtang:
> * What I think is reasonable is a much narrower scope: this layer should primarily serve **frontend IR construction, schedule authoring, and other Allo-specific authoring workflows (as a glue layer)**.

I agree that we can add some Allo-specific wrappers or helpers if that would be very useful. 
If the goal is to serve as a glue layer, does that mean what we actually need is an extra layer on top of the upstream MLIR Python bindings instead of replacing it?


> Provide Allo-specific Python bindings for core upstream MLIR IR concepts:
> 
> * `Context`
> * `Operation`
> * `Block`
> * `Region`

But these seems to go beyond the scope of a glue layer.

Fangtangtang:
> * I looked at the current AIE backend under `allo/backend/aie`, and it currently relies quite heavily on upstream / mlir-aie Python bindings to:
>   
>   * inspect and rewrite IR
>   * build AIE dialect operations
>   * run AIE pass pipelines
>   * generate target-specific IR structures such as tiles, FIFOs, runtime sequences, etc.
> * That dependency is real, but I do not think it argues against the custom bindings themselves. Instead, I think it highlights a natural architectural boundary: target-specific lowering into external dialect ecosystems is better implemented as C++ conversion / lowering passes, not as Python-side IR construction logic.

sad fact, we do not building mlir-aie from source, we are using their pre-built wheels. 


> In other words, for compatibility with out-of-tree dialects, my preferred direction would be:
> * use the custom Python bindings for Allo frontend construction and scheduling
> * implement out-of-tree dialect integration primarily through C++ conversion pipelines
> * expose only a simple Python API on top of those pipelines

you need to build those external dialects with allo to achieve what you want?

---
for migration to a "build together from source" approach, with analysis and passes all implemented in C++. we need to carefully consider the build cost and the developer onboarding overhead

Fangtangtang:
Additionally, I personally feel uneasy about injecting a large number of MLIR-related tools into the Allo project itself (e.g., the `.pyi`), even though they do improve developer experience. If upstream do not have/need them but we really want these tools. I would be more comfortable if these were placed in a separate submodule, so that Allo's core codebase stays focused.

kkkaishao:
> you need to build those external dialects with allo to achieve what you want?

we don't need that. My idea is writing a  "Conversion" pass from Allo to external dialects, then we can use the pass pipelines from other dialects. We do not construct IR for other dialects directly. We only build Allo IR, and then use a C++ conversion pass to lower Allo IR into the corresponding target dialects. This lets us focus only on constructing Allo IR.

This also allows us to use wheels provided by other dialects directly, since they usually include command-line tools. We can invoke those tools seamlessly from Python via subprocess and use their passes without requiring every external dialect to provide Python bindings.

>  I would be more comfortable if these were placed in a separate submodule, so that Allo's core codebase stays focused.

That looks like a good idea🤨, I think that will be a good implementation. Integrating it as a submodule will be better. I also feel that the number of .pyi files is a bit excessive😂.

> Provide Allo-specific Python bindings for core upstream MLIR IR concepts

These concepts are for IR construction. The python bindings are main responsible for IR construction, and serve as a glue layer for Allo-specific authoring workflows (e.g. pass pipeline orchestration) by the way.

> that mean what we actually need is an extra layer on top of the upstream MLIR Python bindings

Yes, that was my original idea as well. But when I started implementing this layer, I realized that we only need a very small subset of the upstream infrastructure. Pulling in the full upstream stack for that feels unnecessarily heavy, and the resulting build artifacts include too many extra pieces. Given that the subset we actually need is fairly small, reimplementing it is not difficult.

> for migration to a "build together from source" approach, with analysis and passes all implemented in C++. we need to carefully consider the build cost and the developer onboarding overhead

Yes, that's the point. I am not sure whether we really need to support Python side analysis and passes. For me the answer is no, and most of passes/analysis are implemented in C++ in MLIR ecosystem, but the answer should depend on our developers' choices.

The module is currently in the following state: it can be integrated as an extension layer on top of upstream, which requires dynamic linking and cannot operate independently of upstream. That approach is viable, although I have not yet tested its stability thoroughly.

It can also be integrated as an all-in-one module, using static linking. In that form, it works fully independently of upstream, and the final artifact is much cleaner. 

The current PR is prepared for the latter setup, but it can also be adapted to support the former one.

As for bringing it into the repository as a git submodule, I agree that this is the right approach (I had forgotten that this was also an option😇).

Fangtangtang:
> My idea is writing a "Conversion" pass from Allo to external dialects, then we can use the pass pipelines from other dialects. We do not construct IR for other dialects directly. We only build Allo IR, and then use a C++ conversion pass to lower Allo IR into the corresponding target dialects. This lets us focus only on constructing Allo IR.

I'm not sure I fully understand how you envision this working.
If the conversion pass needs to emit an MLIR module, I think it would still need to construct operations in the external dialects. (including the headers from the target dialect seems enough, which can be easier than pulling the source and building it together with Allo.)
If the conversion pass just emit string of the module and let the target dialect's tool (either cmd or python binding) parse it, manipulating strings could become quite brittle.

Fangtangtang:
> Yes, that was my original idea as well. But when I started implementing this layer, I realized that we only need a very small subset of the upstream infrastructure. Pulling in the full upstream stack for that feels unnecessarily heavy, and the resulting build artifacts include too many extra pieces. Given that the subset we actually need is fairly small, reimplementing it is not difficult.

I see the motivation for keeping the implementation lightweight. But our requirements may grow over time, and if we reimplement a subset of the upstream infrastructure, we eventually need more functionality that is already provided there. we might end up gradually reimplementing more and more of the upstream stack.

If our current requirements are limited, I still think it makes more sense to keep the glue layer on top of the upstream small. If we later encounter cases where the glue layer is insufficient, we can either extend it with additional features or fall back to the underlying upstream layer.

kkkaishao:
> If the conversion pass needs to emit an MLIR module, I think it would still need to construct operations in the external dialects. (including the headers from the target dialect seems enough, which can be easier than pulling the source and building it together with Allo.)

This is my idea. Certainly manipulating strings are quite brittle and should be avoided as much as possible.

> But our requirements may grow over time, and if we reimplement a subset of the upstream infrastructure, we eventually need more functionality that is already provided there. we might end up gradually reimplementing more and more of the upstream stack.

Well, we are indeed in a period of rapid feature growth. I admit that I am approaching this from the perspective of a feature set that is already relatively stable😂. But I still don't think we'll finally depend on upstream that much. Anyway, it will be better to implement our own subset when our core features are stable enough but not now.

So I think the practical method is:

For the core concepts, such as Operation and Type, we would continue to use the upstream definitions directly.

On top of that, we could expose a set of API bindings whose boundary remains the upstream MLIR objects, while the internal implementation is customized on our side. That would allow our bindings to coexist seamlessly with the upstream bindings. With appropriate boundaries and encapsulation, we can still achieve the interfaces as demonstrated above.

The only exception is `OpBuilder`, which would still need to be bound on our side, since upstream does not provide it.😂 (However InsertionPoint, Location are provided by upstream)

PR #577, opened 2026-04-13
fix: hierarchical dataflow simulator deadlock and HLS codegen (#561, #565)

### LINKED ISSUE #561: [Bug] Hierarchical regions with inter-kernel streams deadlock in simulator and have HLS codegen issues

### LINKED ISSUE #565: [Bug] Missing global_op_cache on ASTContext.copy()

sunwookim028:
Closes #561
Closes #565

## Changes

1. **Simulator recursive OMP** (`allo/backend/simulator.py`): `build_dataflow_simulator`
   only parallelized the top-level function. Inner-region kernel calls ran sequentially,
   deadlocking when kernels communicate via streams. Fix: extracted
   `_inject_omp_parallel_sections()` and applied it recursively to all functions with PE
   kernel calls.

2. **`global_op_cache` missing on copy** (`allo/ir/visitor.py`): `ASTContext.copy()` did
   not copy `global_op_cache`, causing `AttributeError` in nested `@df.kernel` bodies that
   use `@stateful` variables.

3. **HLS codegen: forward declarations + dataflow pragma** (`allo/ir/builder.py`,
   `mlir/lib/Translation/EmitVivadoHLS.cpp`): the emitter emitted inner kernel functions
   after the wrapper that calls them, producing "undeclared identifier" errors in Vitis HLS.
   Fix: added a forward-declaration pass before the full emission pass. Also ensures nested
   `df.region` bodies receive `#pragma HLS dataflow`. `nameTable` is cleared between passes
   so type names are re-emitted correctly.
   Note: cherry-picked onto current main after #557's `emitFunctionSignature` refactor;
   conflict resolved by preserving #557's `SmallVector<Value,8>` return type while
   grafting in the forward-declaration loop and `nameTable.clear()`.

## Tests
- `tests/dataflow/test_hierachical.py` — simulator: **PASSED**
- HLS codegen tests require Vitis HLS (brg-zhang-xcel only)

### DIFF
--- a/allo/backend/simulator.py
+++ b/allo/backend/simulator.py
@@ -79,6 +79,7 @@ def _process_function_streams(
     module: Module,
     func: func_d.FuncOp,
     processed_funcs: set,
+    all_pe_calls_by_func: dict = None,
 ):
     """
     Process streams and PE calls within a single function.
@@ -109,7 +110,7 @@ def _process_function_streams(
                         if callee_name == str(mod_op.sym_name).strip('"'):
                             pe_call_define_ops[op] = mod_op
                             # Recursively process the callee function first
-                            _process_function_streams(module, mod_op, processed_funcs)
+                            _process_function_streams(module, mod_op, processed_funcs, all_pe_calls_by_func)
                             break
         elif isinstance(op, allo_d.StreamConstructOp):
             stream_name = str(op.attributes["name"]).strip('"')
@@ -765,12 +766,42 @@ def _process_function_streams(
     for op in stream_construct_ops.values():
         op.operation.erase()
 
+    # Accumulate PE calls keyed by function for recursive OMP injection
+    if all_pe_calls_by_func is not None and pe_call_define_ops:
+        all_pe_calls_by_func[func_name] = pe_call_define_ops
+
     return (
         stream_struct_table,
         stream_type_table,
         pe_call_define_ops,
         stream_construct_ops,
     )
+
+
+def _inject_omp_parallel_sections(pe_call_define_ops):
+    """Wrap a set of func.call ops in omp.parallel > omp.sections > omp.section blocks."""
+    assert len(pe_call_define_ops) > 0
+    omp_ip = InsertionPoint(beforeOperation=list(pe_call_define_ops.keys())[0])
+    omp_parallel_op = openmp_d.ParallelOp([], [], [], [], ip=omp_ip)
+    assert isinstance(omp_parallel_op.region, Region)
+    omp_parallel_block = Block.create_at_start(omp_parallel_op.region, [])
+
+    # Add `omp.sections`
+    ip_omp_parallel = InsertionPoint(omp_parallel_block)
+    omp_sections_op = openmp_d.SectionsOp([], [], [], [], ip=ip_omp_parallel)
+    omp_sections_block = Block.create_at_start(omp_sections_op.region, [])
+    openmp_d.TerminatorOp(ip=ip_omp_parallel)
+
+    # Add `omp.section`s for PE calls
+    ip_omp_sections = InsertionPoint(omp_sections_block)
+    for call_op in pe_call_define_ops:
+        assert isinstance(call_op, OpView)
+        omp_section_op = openmp_d.SectionOp(ip=ip_omp_sections)
+        omp_section_block = Block.create_at_start(omp_section_op.region, [])
+        ip_omp_section = InsertionPoint(omp_section_block)
+        omp_term_op = openmp_d.TerminatorOp(ip=ip_omp_section)
+        call_op.operation.move_before(omp_term_op.operation)
+    openmp_d.TerminatorOp(ip=ip_omp_sections)
 
 
 def build_dataflow_simulator(module: Module, top_func_name: str):
@@ -799,12 +830,13 @@ def build_dataflow_simulator(module: Module, top_func_name: str):
 
         # Process all functions with streams recursively, starting from top
         processed_funcs: set = set()
+        all_pe_calls_by_func: dict = {}
         func = find_func_in_module(module, top_func_name)
         assert isinstance(func.body, Region)
 
         # Recursively process the top function and all its callees
         _, _, pe_call_define_ops, _ = _process_function_streams(
-            module, func, processed_funcs
+            module, func, processed_funcs, all_pe_calls_by_func
         )
 
         # If no PE calls were found in top function, collect them again from the processed functions
@@ -819,30 +851,12 @@ def build_dataflow_simulator(module: Module, top_func_name: str):
                                 if callee_name == str(mod_op.sym_name).strip('"'):
                                     pe_call_define_ops[op] = mod_op
                                     break
+            all_pe_calls_by_func[top_func_name] = pe_call_define_ops
 
-        # Add the outmost `omp.parallel`
-        assert len(pe_call_define_ops) > 0
-        omp_ip = InsertionPoint(beforeOperation=list(pe_call_define_ops.keys())[0])
-        omp_parallel_op = openmp_d.ParallelOp([], [], [], [], ip=omp_ip)
-        assert isinstance(omp_parallel_op.region, Region)
-        omp_parallel_block = Block.create_at_start(omp_parallel_op.region, [])
-
-        # Add `omp.sections`
-        ip_omp_parallel = InsertionPoint(omp_parallel_block)
-        omp_sections_op = openmp_d.SectionsOp([], [], [], [], ip=ip_omp_parallel)
-        omp_sections_block = Block.create_at_start(omp_sections_op.region, [])
-        openmp_d.TerminatorOp(ip=ip_omp_parallel)
-
-        # Add `omp.section`s for PE calls
-        ip_omp_sections = InsertionPoint(omp_sections_block)
-        for call_op in pe_call_define_ops:
-            assert isinstance(call_op, OpView)
-            omp_section_op = openmp_d.SectionOp(ip=ip_omp_sections)
-            omp_section_block = Block.create_at_start(omp_section_op.region, [])
-            ip_omp_section = InsertionPoint(omp_section_block)
-            omp_term_op = openmp_d.TerminatorOp(ip=ip_omp_section)
-            call_op.operation.move_before(omp_term_op.operation)
-        openmp_d.TerminatorOp(ip=ip_omp_sections)
+        # Inject omp.parallel/sections into every function that has PE calls
+        for func_pe_calls in all_pe_calls_by_func.values():
+            if func_pe_calls:
+                _inject_omp_parallel_sections(func_pe_calls)
 
 
 # This pass is only meant to run on fully lowered MLIR code
--- a/allo/backend/simulator.py
+++ b/allo/backend/simulator.py
@@ -110,7 +110,9 @@ def _process_function_streams(
                         if callee_name == str(mod_op.sym_name).strip('"'):
                             pe_call_define_ops[op] = mod_op
                             # Recursively process the callee function first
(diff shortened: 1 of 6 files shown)


mlir/lib/Translation/EmitVivadoHLS.cpp
Fangtangtang, line 1: does #557 still fail to fix the forward reference issue?

allo/ir/builder.py
```diff
@@ -2729,13 +2739,22 @@ def build_Call(ctx: ASTContext, node: ast.Call, out_buffer: OpView = None):
                     new_ctx.func_suffix = inst_suffix
 
             func_op = ASTTransformer.build_FunctionDef(new_ctx, func_def)
+            func_op.attributes["dataflow"] = UnitAttr.get()
+            if ctx.top_func is not None:
+                func_op.operation.move_before(ctx.top_func)
```
Fangtangtang, line 2744: I guess this is also added to fix foward reference. Forward reference is valid in mlir but invalide in HLS backend. We'd better fix the issue in HLS backend codegen.
Could you please check whether existing fix (introduced in #557 ) already works?

allo/ir/builder.py
```diff
@@ -2729,13 +2739,22 @@ def build_Call(ctx: ASTContext, node: ast.Call, out_buffer: OpView = None):
...
             new_args = build_stmts(ctx, node.args)
-            arg_values = [
-                ASTTransformer.get_mlir_op_result(ctx, arg) for arg in new_args
-            ]
+            arg_values = []
+            for arg in new_args:
+                res = ASTTransformer.get_mlir_op_result(ctx, arg)
+                if isinstance(res.type, MemRefType) and len(res.type.shape) == 0:
+                    op_ = ASTTransformer.build_scalar(ctx, arg)
+                    arg_values.append(op_.result)
```
Fangtangtang, line 2754: this would pass the scalar argument by value. Is this the intended behavior?

allo/ir/builder.py
```diff
@@ -2211,10 +2211,20 @@ def build_FunctionDef(ctx: ASTContext, node: ast.FunctionDef):
...
+                                        res = ASTTransformer.get_mlir_op_result(
+                                            ctx, arg
+                                        )
+                                        # If it's a 0D memref (scalar), load it to get the value
+                                        if (
+                                            isinstance(res.type, MemRefType)
+                                            and len(res.type.shape) == 0
+                                        ):
+                                            op_ = ASTTransformer.build_scalar(ctx, arg)
+                                            arg_values.append(op_.result)
```
Fangtangtang, line 2225: this would pass the scalar argument by value.
Also, I think `args` should only contain tensors (please correct me if I'm missing something). In that case, we should raise an error in the type infer phase if the user use a scalar here.

allo/ir/builder.py
Fangtangtang, line 1994: can we remove this since `hasattr(ctx, "global_op_cache")` is always `True` after the update in `ASTContext`

allo/backend/simulator.py
```diff
@@ -765,6 +768,10 @@ def _process_function_streams(
     for op in stream_construct_ops.values():
         op.operation.erase()
 
+    # Accumulate PE calls keyed by function for recursive OMP injection
+    if all_pe_calls_by_func is not None and pe_call_define_ops:
+        all_pe_calls_by_func[func_name] = pe_call_define_ops
```
Fangtangtang, line 773: always not `None`?

Fangtangtang (commented):
Thanks for the fixes! I just have some suggestions.

Also, could you add some tests to cover these fixes? It looks like they are addressing issues in hierarchical designs.

sunwookim028:
Thanks for the review @Fangtangtang! Addressed three of the comments and have questions on the rest.

**Pushed in 7a3212c:**

- `builder.py:1994` — removed the `hasattr(ctx, "global_op_cache")` guard. With `visitor.py` initializing it in `__init__` and propagating in `copy()`, the check is dead code. Good catch.
- `simulator.py:773` — made `all_pe_calls_by_func` a required parameter (every caller already provides one) and dropped the `is not None` guard.
- `EmitVivadoHLS.cpp` — you were right: #557's `emitFunctionDeclaration()` already emits selective forward declarations and properly saves/restores `state.nameTable` / `state.nameConflictCnt`. My added Third-pass was duplicating that work and clobbering emitter state. Removed the pass and the `nameTable.clear()` workaround.

**Want your guidance before I touch the rest:**

- `builder.py:2744` — `func_op.operation.move_before(ctx.top_func)` in `build_Call`. Same forward-reference theme: this moves the callee's `func.func` op above the caller's. With #557's HLS-side selective forward decls handling out-of-order calls in the emitted C++, do you think the IR-level move is still needed for any backend (simulator / passes.py reorder pass)? I suspect it's now redundant but I haven't bisected it against the original `test_hierachical.py` repros.

- `builder.py:2225` and `:2754` (scalar-by-value 0D-memref handling) — agree that `args=[…]` ideally only contains tensors. The 0D-memref case showed up because region-scope scalar parameters become 0D memrefs in MLIR. Two options I see: (1) raise a type-infer error if a scalar appears in `args`, and require users to pass a 1-element tensor; or (2) keep the load-and-pass-by-value behavior as a convenience. Which would you prefer? If (1), I'll add the check in the type-infer phase as you suggested.

- Tests: I'll add a hierarchical-dataflow regression test exercising the simulator OMP fix once the above two questions settle, since the test shape depends on which path we keep.

Fangtangtang:
Thanks! Sorry for the delayed response.

> * `builder.py:2744` — `func_op.operation.move_before(ctx.top_func)` in `build_Call`.

I think we don't need this at IR level.  Forward reference is valid in mlir and some mlir based backend. 

> * `builder.py:2225` and `:2754` (scalar-by-value 0D-memref handling) 

I prefer (1).

mlir/lib/Translation/EmitVivadoHLS.cpp
Fangtangtang: shall we just restore this file?

Fangtangtang (commented):
mostly LGTM! just one minor suggestion.

chhzh123:
> mostly LGTM! just one minor suggestion.

@Fangtangtang Feel free to merge the PRs if you have reviewed them

sunwookim028:
Good catch — you're right. The `portList.push_back` reorder and comment rewording were incidental cleanup that rode in alongside the Third-pass removal in 7a3212c. Since `portList` is only consumed after `emitFunctionSignature` returns (by `emitFunctionDirectives`), the reorder is a pure no-op. Restored `EmitVivadoHLS.cpp` to the upstream baseline in b5ee250. The functional fixes for #561/#565 live entirely in `builder.py`, `simulator.py`, and `passes.py`. Ready for re-review!

Fangtangtang (commented):
Great work! Could you fix the pylint warning (R0912)? I'll merge this once the CI is green.

sunwookim028:
@Fangtangtang done! I have few other PR/issue planned regarding hierarchical HLS designs. Should we discuss them as issues/PR/discussion, or just offline?

Fangtangtang:
Thanks!

> I have few other PR/issue planned regarding hierarchical HLS designs. Should we discuss them as issues/PR/discussion, or just offline?

Issues/PRs/discussions would probably be best so others can also follow the design discussion and future changes. But happy to chat offline first if that's easier for brainstorming.

sunwookim028:
Sounds good. I'll just use those features then!

PR #579, opened 2026-04-14
fix: fp16 (half) type support in Vitis HLS backend

sunwookim028:
## Problem

Two bugs prevented `float16` kernels from synthesizing correctly with Vitis HLS:

1. **Builder type guard** (`allo/ir/builder.py`): the scalar math dispatch
   check only accepted `F32Type` and `F64Type`.  `float16` operands fell
   through, so `allo.exp(float16)` was not lowered to `math.ExpOp`.

2. **HLS emitter namespace** (`mlir/lib/Translation/EmitVivadoHLS.cpp`):
   for `Float16Type` operands, unary math calls (exp, log, sqrt, sin, cos,
   tanh, exp2, log2, log10, abs) must use the `hls::` prefix.
   `hls_math.h` places the `half` overloads inside `namespace hls`; the
   bare calls are ambiguous with the C `double` / C++ `float` overloads
   and cause csyn to fail.

## Changes

- `allo/ir/builder.py`: add `F16Type` and `BF16Type` to the scalar math
  dispatch type guard.
- `mlir/lib/Translation/EmitVivadoHLS.cpp`: add `isHalf()` helper; emit
  `hls::<fn>` instead of `<fn>` for all scalar math unary ops when the
  operand type is `Float16Type`.

## Verification

Both fixes verified with Vitis HLS 2023.2 csyn on U280 (411 MHz target):

| Kernel | LUT | FF | BRAM | DSP | II |
|---|---|---|---|---|---|
| fp16 add+scale | 3986 | 4233 | 6 | 5 | 21 |
| fp16 exp | 2668 | 2576 | 4 | 2 | 13 |


mlir/lib/Translation/EmitVivadoHLS.cpp
```diff
@@ -528,23 +528,49 @@ class ExprVisitor : public HLSCppVisitorBase<ExprVisitor, bool> {
...
+  // For half (Float16Type) operands, unqualified math calls are ambiguous
+  // between C double and C++ float overloads.  hls_math.h declares these
+  // inside namespace hls, so we use the hls:: prefix to select the half
+  // overload unambiguously.
+  static bool isHalf(Operation *op) {
+    return llvm::isa<Float16Type>(op->getOperand(0).getType());
+  }
+  bool visitOp(math::AbsFOp op) {
+    return emitter.emitUnary(op, isHalf(op) ? "hls::abs" : "abs"), true;
+  }
```
Fangtangtang, line 540: can we use `hls::xxx` for all the data types we currently support?

Fangtangtang (commented):
Thanks! Can you add some test cases for the new feature?

sunwookim028:
Thanks @Fangtangtang! Pushed a9efbea:

- **EmitVivadoHLS.cpp** — applied your suggestion: switched every FP math op (sin/cos/tanh/sqrt/rsqrt/exp/exp2/log/log2/log10/abs/ceil) to use the `hls::` prefix unconditionally. Dropped the half-only `isHalf()` helper. Rationale: `hls_math.h` provides overloads for half/float/double/ap_fixed, so the qualified form is unambiguous for half and recommended by Vitis HLS for the others.
- **Tests** — added `tests/test_vhls.py::test_fp16_half_type_and_hls_math`, which builds a `float16[8]` kernel using `allo.exp` + `allo.sqrt` and checks that the generated HLS contains `half`, `hls::exp`, and `hls::sqrt`. Lets me know if you'd like additional coverage (e.g., fp16 + ExternalModule path, or a csim run gated on `hls.is_available`).

Fangtangtang:
Could you please fix the formatting issue?
You can run the format checker locally with: `bash scripts/lint/task_lint.sh`

sunwookim028:
Fixed — the `hls::` prefixed one-liner visitOp bodies in `EmitVivadoHLS.cpp` exceeded the column limit after the hls:: prefix was added. Applied clang-format in 2013299; lint passes cleanly now.

tests/test_vhls.py
```diff
@@ -639,5 +639,21 @@ def floordiv(A: int32[10], B: int32[10]) -> int32[10]:
         assert " / " in hls_code
 
 
+def test_fp16_half_type_and_hls_math():
+    def kernel(A: float16[8]) -> float16[8]:
```
Fangtangtang: The unit test failed due to a conflict in the kernel name. 
I can push a small fix to your branch if you're okay with it 👀

sunwookim028: Oh thanks! That'll be awesome!

PR #600, opened 2026-07-25
[Fix][Dataflow] Fix Argument Mapping in Top Module

### LINKED ISSUE #592: [Bug] @df.region merges top-level arguments by kernel local parameter name → distinct region inputs  silently aliased

### LINKED ISSUE #596: [Bug][Dataflow][IR] Incorrect Argument Mapping Based on Appearance Order
### Describe the bug

Current argument mapping logic relies on their appearance order. It should be based on names instead.

### Reproduction

This issue leads to the bug described in #592. The fix in #593 partially patches #592, but it does not address the issue in the argument mapping logic.

The following test case shows that bug still exists even after applying #593:
```python
import allo
from allo.ir.types import int32
import allo.dataflow as df
import numpy as np


@df.region()
def top(a: int32[4], b: int32[4]):
    @df.kernel(mapping=[1], args=[b])
    def k_a(x: int32[4]):
        for i in range(4):
            x[i] = 1

    @df.kernel(mapping=[1], args=[a])
    def k_b(x: int32[4]):
        for i in range(4):
            x[i] = 2


if __name__ == "__main__":
    mod = df.build(top, target="simulator")
    a = np.zeros(4, dtype=np.int32)
    b = np.zeros(4, dtype=np.int32)
    mod(a, b)
    print("a:", a, " b:", b)
```

### Buggy output

output:
```bash
a: [1 1 1 1]  b: [2 2 2 2]
```

### Expected behavior

expected output:
```bash
a: [2 2 2 2]  b: [1 1 1 1]
```

### Additional context

_No response_

Fangtangtang:
## Description ##
Fixes #592 and fixes #596.

**Known limitation**: This fix does not consider hierarchical design. In hierarchical design, the argument mapping for `df.region` follows a different implementation logic, and handling this requires careful consideration of the IR builder design. This will be addressed in the future.

## Checklist ##

- [ ] Code is well-documented

### DIFF
--- a/tests/dataflow/aie/test_matrix.py
+++ b/tests/dataflow/aie/test_matrix.py
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import allo
-from allo.ir.types import int16, int32
+from allo.ir.types import int16, int32, Stream
 import allo.dataflow as df
 import numpy as np
 from allo.memory import Layout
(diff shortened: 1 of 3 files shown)


tests/dataflow/aie/test_matrix.py
```diff
@@ -185,12 +185,57 @@ def gemm(
...
+        @df.kernel(mapping=[1], args=[C])
+        def store(local_C: Ty[M, N]):
+            local_C[:Ma, :] = pipeC[0].get()
+            local_C[Ma:, :] = pipeC[1].get()
+
+    A = np.random.randint(0, 64, (M, K)).astype(np.int16)
+    B = np.random.randint(0, 64, (K, N)).astype(np.int16)
+    C = np.zeros((M, N)).astype(np.int16)
+    if is_available():
+        mod = df.build(top, target="aie")
```
Fangtangtang, line 224: typo: should be `top1`. will fix in next PR

PR #610, opened 2026-09-08
Limit C linkage to HLS top functions

### LINKED ISSUE #603: [Bug] extern "C" wraps helper functions returning non-POD HLS types (ap_int<N>), breaking compilation
### Describe the bug

`postprocess_hls_code()` in the vitis_hls backend wraps every function in the generated kernel.cpp — from `using namespace std;` down to EOF — inside a single `extern "C" { ... }` block, without distinguishing the top function from internal helper functions. 
When a kernel calls a helper function that returns a custom-bitwidth type (e.g. `Int(22)`, emitted as `ap_int<22>`), that helper ends up with C linkage on a non-POD C++ class return type. 
Vitis HLS's csynth front-end reports this as a hard error
(`HLS 207-5153`), so `csynth_design` fails and the whole build aborts.

### Reproduction

```python
  import allo
  from allo.ir.types import int32, Int

  def helper_func(x: int32) -> Int(22):
      return x * 2

  def main_func(arr: int32[10]) -> Int(22)[10]:
      result: Int(22)[10]
      for i in allo.grid(10):
          result[i] = helper_func(arr[i])
      return result

  s = allo.customize(main_func)
  mod = s.build(target="vitis_hls", mode="csyn", project="csyn.prj")
  mod()  # runs `vitis_hls -f run.tcl`, i.e. actual C synthesis
```

### Environment
- Vitis HLS: 2023.2.2

### Buggy output

Generated `csyn.prj/kernel.cpp` (real output of the backend):

```cpp
  using namespace std;

  extern "C" {

  ap_int<22> helper_func(
    int32_t v0
  ) {
    return ((ap_int<22>)((((int64_t)(v0)) * 2)));
  }

  void main_func(
    int32_t *v9,
    ap_int<22> *v10
  ) {
    ...
    result[i] = helper_func(v14);
    ...
  }

  } // extern "C"
```

Running `mod()` invokes `vitis_hls -f run.tcl`, which fails during
`csynth_design`:

```bash
  WARNING: [HLS 207-5153] 'helper_func' has C-linkage specified, but returns
  incomplete type 'ap_int<22>' which could be incompatible with C
  (kernel.cpp:20:12)
  ERROR: [HLS 207-5153] 'helper_func' has C-linkage specified, but returns
  incomplete type 'ap_int<22>' which could be incompatible with C
  (kernel.cpp:20:12)
  Compilation of the preprocessed source 'kernel' failed
```

which surfaces in Python as:

  RuntimeError: Failed to synthesize the design

Removing the `extern "C" { ... }` wrapper around `helper_func` (leaving it
around `main_func` only) makes the same file synthesize without this error.

### Expected behavior

`csynth_design` should succeed. Only the top function needs C linkage (so it can be called from the OpenCL/host runtime); internal helper functions, especially ones returning non-POD HLS types (`ap_int`/`ap_uint`/`ap_fixed`), should not be placed inside `extern "C"`.

### Additional context

_No response_

appleweiping:
Fixes #603.

The generated C linkage now applies only to the top HLS function. Internal helpers retain C++ linkage, so helpers returning types such as ap_int are no longer placed inside a global extern C block.

The same correction is applied to the Vitis and PYNQ postprocessors. The new regression tests cover a helper returning ap_int and verify that only the top function receives C linkage.

Verification:

- Python compilation passed for both postprocessors and the new test
- The isolated postprocessor regression passed for both backends
- Black 24.8.0 check passed
- License header check passed

Vitis HLS is not available on this machine, so hardware synthesis was not run.


Fangtangtang (changes requested):
Thanks for the PR! Could you confirm whether you have run all the relevant test cases on the backend affected by this change, and share the results?

I'm asking because the proposed fix doesn't quite make sense to me, so I'd like to see more evidence.

appleweiping:
Thanks for checking. On this commit, the self-hosted `Allo Test` workflow built Allo and ran the full repository suites successfully:

- non-dataflow suite: 500 passed, 42 skipped
- dataflow suite: 30 passed, 3 skipped
- affected backend coverage within that run: `test_hls_postprocess.py` 3 passed, `test_pynq.py` 2 passed, `test_vitis.py` 11 passed, and `test_vhls.py` 25 passed

Run: https://github.com/cornell-zhang/allo/actions/runs/34191910780/job/101951527524

The intended change keeps C linkage on the exported kernel. The generated top definition changes from being inside a file-wide `extern "C"` block to `extern "C" void <top>(...)`, while helper definitions keep C++ linkage. This is specifically to avoid Vitis HLS 207-5153 when a helper returns `ap_int<N>`.

I do not have Vitis HLS installed locally, so I have not independently rerun `csynth_design`; the workflow evidence above covers code generation and the existing Vitis, VHLS, and PYNQ test suites, but not actual Vitis synthesis. If useful, I can also add an end-to-end Allo code-generation regression based directly on the reproducer in issue 603.

appleweiping:
I added the code generation regression from issue 603 in 9a8a864. It verifies that the generated helper keeps C++ linkage and uses an ap_int<22> output while only the top function has C linkage. The full workflow is green: 501 tests passed in the main suite and 30 passed in the dataflow suite. The accelerator, tutorial, benchmark, MachSuite, and PyTorch steps also passed. Please re-review when convenient.

